PR #2, opened 2024-06-21
Jinja2

RealOrangeOne:
I've taken a few sensible defaults, but we can easily tweak these if we need something else when integrating the DS.


zerolab:
Note that this needs https://service-manual.ons.gov.uk/design-system/guidance/templating-in-nunjucks#using-nunjucks-in-jinja2-templates in prep for DS integration

zerolab (approved):
The cleanup looks sane to me (but I didn't test this locally)

PR #6, opened 2024-06-27
Integrate the Design System

zerolab:
This PR:

- drops all node dependencies
- adds a Makefile to pull in the DS templates into `ons_alpha/jinja2`
- moves existing templates to `ons_alpha/jinja2/templates` to avoid colisions with the DS ones
- updates settings and base templates to integrate with it

To-Do

- [x] add `make load-design-system-templates` to the image build
- [x] finish the other templates (eg. password required)
- [ ] plug in nav (header/footer) - in another PR
- [x] plug in breadcrumbs

Screencast: https://www.loom.com/share/816a9970d98e4eecbb847f99e152a7a3?sid=17fd9a94-8222-465f-b226-8f4540a0db84


ons_alpha/jinja2/templates/base.html
```diff
@@ -0,0 +1,59 @@
...
+}
+],
+"OGLLink": {
+"pre": 'All content is available under the',
+"link": 'Open Government Licence v3.0',
+"url": 'https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/',
+"post": ', except where otherwise stated'
+}
+}
+} %}
```
MebinAbraham: Minor, djlint doesn't play well with jinja macros but given djlint is disabled for this block can it be manually formatted. It makes it hard to read as is.

I'm happy to leave the other small macro calls as is. We can sort it all out for beta build.

.gitignore
```diff
@@ -46,3 +46,8 @@ coverage.xml
 
 /database_dumps
 dpl-rendered-patterns
+
+/ons_alpha/jinja2/assets/styles/print.css
+/ons_alpha/jinja2/components
+/ons_alpha/jinja2/layout
```
MebinAbraham, line 52: This should be added to `.dockerignore` as well.

Makefile
```diff
@@ -0,0 +1,13 @@
...
+
+.DEFAULT_GOAL := help
+
+.PHONY: help load-design-system-templates
+
+help:  ## ⁉️   - Display help comments for each make command
+	@awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z0-9_\-\.]+:.*?## / {printf "  %-20s %s\n", $$1, $$2}' $(MAKEFILE_LIST)
+
+
+load-design-system-templates:  ## 🎨️   - Load the design system templates
```
MebinAbraham, line 11: This needs to be called in the Dockefile. Currently, it is only working if I run this locally then build the image (which doesn't exclude the locally sourced templates).

MebinAbraham (commented):
The `docs/front-end` can be removed since all the node tooling has been removed.

Only done a quick pass. There are quite a few things we would need to change for beta but for alpha, it LGTM.

Just raising for visibility and not for this PR, we should make the alpha repo even more minimal, it has a bunch of files/config that are not required leading to unnecessary confusion.

ons_alpha/jinja2/templates/components/streamfield/call_to_action_block.html
```diff
@@ -0,0 +1,3 @@
+{% with call_to_action=value %}
+    {% include "components/cta/call_to_action.html" %}
```
MebinAbraham: Broken reference.
```suggestion
    {% include "templates/components/cta/call-to-action.html" %}
```

ons_alpha/jinja2/templates/components/cta/call-to-action.html
```diff
@@ -1,5 +1,3 @@
-{% load wagtailcore_tags wagtailimages_tags %}
-
 {% comment %}
```
MebinAbraham: This currently errors, this tag need to be change to [Jinja comments](https://jinja.palletsprojects.com/en/3.1.x/templates/#comments). `{% comment %}` is Django.

ons_alpha/jinja2/templates/components/cta/call-to-action.html
```diff
@@ -15,9 +13,9 @@
 
         <h2>{{ call_to_action.title }}</h2>
 
-        {{ call_to_action.summary|richtext }}
+        {{ call_to_action.summary|richtext() }}
 
-        {% image call_to_action.image fill-450x300 %}
+        {% image(call_to_action.image, "fill-450x300") %}
 
         <p><a href="{{ call_to_action.get_link_url }}">{{ call_to_action.get_link_text }}</a></p>
```
MebinAbraham: Image rendering seem to be broken and the CTA link doesn't render correctly either.

ons_alpha/jinja2/templates/base.html
zerolab: Just realised we use djhtml which is why the formatting was off. updated

MebinAbraham (commented):
~~The DS templates are still not available within the container. I get `TemplateNotFound` errors. I inspected the container after the build and the DS templates are not in `jinja2/`~~

EDIT: Sorted, was due to cache.

MebinAbraham (approved):

MebinAbraham (approved):

PR #7, opened 2024-07-01
Bulletin pages

RealOrangeOne:
https://jira.ons.gov.uk/browse/NWP-145

Adds a skeleton bulletin page (with related topic and topic section pages).

The content is mostly placeholders to be used to build out content. Basic blocks exist. The notably missing areas are "Related links", Citations, table downloads, charts.


ons_alpha/jinja2/templates/components/streamfield/stream_block.html
```diff
@@ -1,3 +0,0 @@
-{% for block in value %}
```
zerolab: please leave this as I am using it in #8

ons_alpha/jinja2/templates/pages/topics/topic_section_page.html
```diff
@@ -0,0 +1 @@
+{% extends "templates/base_page.html" %}
```
zerolab: please re-use the standard index template. it has a good enough starting point

ons_alpha/jinja2/templates/components/streamfield/stream_block.html
RealOrangeOne: We've gone in different directions. I've used a list of blocks and used `include_blocks`, whilst you've used a struct block to store which blocks are valid. I'm not sure what the best implementation is?

zerolab: Both implementations are valid. I am more in favour of defining `StreamBlock`s as usually we have more than a few block types

RealOrangeOne: I guess the main benefit of that approach is it lets us reuse the same block set on multiple models more easily. I'll port mine.

RealOrangeOne: I've swapped to using a `StreamBlock`, but even then this template isn't necessary - looping through and rendering each block is the default, and removing the template removes the jinja overhead.

ons_alpha/jinja2/templates/components/streamfield/panel_block.html
```diff
@@ -1,7 +1,7 @@
 {% from "components/panel/_macro.njk" import onsPanel %}
 
 {{ onsPanel({
-    "variant": value.variant,
-    "body": value.body,
-    "title": value.title
+"variant": value.variant,
```
zerolab: you can wrap these with `{# fmt:off #}` / `{# fmt:on #}` to prevent djhtml messing them up

RealOrangeOne: `djhtml` has an oddly hard time with funky indentation. Yeah that makes things much clearer!

ons_alpha/bulletins/blocks.py
```diff
@@ -0,0 +1,14 @@
...
+    PanelBlock,
+    TableBlock,
+)
+
+
+class BulletinStoryBlock(StreamBlock):
+    heading = HeadingBlock()
+    rich_text = RichTextBlock()
+    panel = PanelBlock()
+    table = TableBlock()
```
zerolab, line 16: bulletins do use images, as well as embedded videos at times, so `ImageBlock`/`EmbedBlock` from the old core blocks would be useful here.

Looking at this.. I want us to keep the information page and the index page models, even if they are not strictly part of the the thin slice

ons_alpha/core/blocks/markup.py
```diff
@@ -0,0 +1,45 @@
+from django.utils.text import slugify
+from wagtail import blocks
+from wagtail.contrib.table_block.blocks import TableBlock as WagtailTableBlock
+
+
+class HeadingBlock(blocks.CharBlock):
+    class Meta:
+        template = "templates/components/streamfield/heading_block.html"
```
zerolab, line 13:
```suggestion
        icon = "title"
        template = "templates/components/streamfield/heading_block.html"
```

ons_alpha/core/blocks/panels.py
```diff
@@ -0,0 +1,21 @@
...
+            ("success", "Success"),
+            ("warn-branded", "Warn (branded)"),
+            ("warn", "Warn"),
+        ]
+    )
+    title = blocks.CharBlock(required=False)
+    body = blocks.RichTextBlock()
+
+    class Meta:
+        template = "templates/components/streamfield/panel_block.html"
```
zerolab, line 21: note: best to add icons for all SF blocks.. makes for a nicer editorial experience (and our goal is to delight). can be done as a follow up when we add related links, for example

ons_alpha/core/models/mixins.py
```diff
@@ -76,3 +78,21 @@ class Meta:
             ],
         )
     ]
+
+
+class SubpageMixin:
+    PAGE_SIZE = 24
```
zerolab, line 84: question: why 24? :)
and should this not be configurable via settings?

ons_alpha/core/templatetags/tableofcontents.py
```diff
@@ -0,0 +1,18 @@
+import jinja2
+
+from django_jinja import library
+
+
+@library.global_function
+@library.render_with("templates/components/navigation/table-of-contents.html")
+@jinja2.pass_context
+def table_of_contents(context, page, attr_name, options=None):
```
zerolab: note to self: in a future piece of work we should rename this to `table_of_contents_from_streamfield`

also looking at https://service-manual.ons.gov.uk/design-system/components/table-of-contents we could probably use the "In-page table of contents" variant where `{"lists": [{"itemsList": ...}` is replaced by `{"itemsList":...}`

zerolab (changes requested):

ons_alpha/core/models/mixins.py
RealOrangeOne: The designs I saw used 3 columns, so 24 gives a reasonable number of cards before paginating.

It's configurable per-page by just setting `PAGE_SIZE` on it. I wouldn't say it needs to be a Django setting.

zerolab:
Needs a rebase on main. and a `make format`

zerolab (approved):
Will merge this, then follow up with a PR for a few of the missing elements.

PR #12, opened 2024-07-04
Bulletin page enhancements

zerolab:
https://jira.ons.gov.uk/browse/NWP-198

Follows from #7 and

- adds a field for corrections/updates ([example](https://www.ons.gov.uk/businessindustryandtrade/business/businessservices/bulletins/nonfinancialbusinesseconomyukannualbusinesssurvey/2022results)) -- I suspect the main block title should vary depending on whether it has only correction, only updates or both
- is_accredited (the PSF bulletins show the accreditaion logo in the content header)
- contact details (when wiring in the templates, should be in the content header, and the sidebar)
- updates the body StreamField with
  - related links 
  - ONS embed (requires a URL starting with `https://www.ons.gov.uk/visualisations/`. e.g. https://www.ons.gov.uk/visualisations/dvc2975/dashwrapper/index.html)
  - a MathJax block

Does some very light templating, but would be good to get Nick to wire in the templates

PR #14, opened 2024-07-10
Feature/bulletin page UI improvements

mr-nick17:
### What

Bulletin page enhancements
- Mapped components to typical layout with the ONS design system, using newer web services as a guide (WIP/best endeavours) 
- Adding all toc elements
- fixed missing markup on related links

### How to review

Sense check
Image review

image

### Who can review

!me

### DIFF
--- a/ons_alpha/bulletins/models.py
+++ b/ons_alpha/bulletins/models.py
@@ -1,3 +1,4 @@
+from functools import cached_property
 from django.db import models
 from wagtail.admin.panels import FieldPanel, FieldRowPanel, MultiFieldPanel, ObjectList, TabbedInterface
 
@@ -58,3 +59,23 @@ class BulletinPage(BasePage):
             ObjectList(BasePage.settings_panels, heading="Settings", classname="settings"),
         ]
     )
+
+    def get_context(self, request, *args, **kwargs):
+        context = super().get_context(request, *args, **kwargs)
+        
+        context["toc"] = self.toc
+
+        return context
+
+    @cached_property
+    def toc(self):
+        items = [{"url": "#summary", "text": "Summary"}]
+        
+        for block in self.body:  # pylint: disable=not-an-iterable
+            if hasattr(block.block, "to_table_of_contents_items"):
+                items += block.block.to_table_of_contents_items(block.value)
+
+        if self.contact_details_id:
+            items += [{"url": "#contact-details", "text": "Contact details"}]
+
+        return items
--- a/ons_alpha/bulletins/models.py
+++ b/ons_alpha/bulletins/models.py
@@ -1,4 +1,5 @@
 from functools import cached_property
+
 from django.db import models
 from wagtail.admin.panels import FieldPanel, FieldRowPanel, MultiFieldPanel, ObjectList, TabbedInterface
 
@@ -62,7 +63,7 @@ class BulletinPage(BasePage):
 
     def get_context(self, request, *args, **kwargs):
         context = super().get_context(request, *args, **kwargs)
-        
+
         context["toc"] = self.toc
 
         return context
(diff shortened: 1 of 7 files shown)


zerolab:
#15 for adding related links to the TOC

We will want to improve the toc tag to allow arbitrary elements to be added it to it, but that is for another task (if needed)

zerolab:
@mr-nick17 #12 added an ONS embed block which wraps the provided in an iframe that is currently unstyled

mr-nick17:
> #15 for adding related links to the TOC
> 
> We will want to improve the toc tag to allow arbitrary elements to be added it to it, but that is for another task (if needed)

Yep agree, would be great to understand how that works and/or how to add items that aren't on the `page.body`

ons_alpha/bulletins/models.py
```diff
@@ -58,3 +60,23 @@ class BulletinPage(BasePage):
...
+    @cached_property
+    def toc(self):
+        items = [{"url": "#summary", "text": "Summary"}]
+
+        for block in self.body:  # pylint: disable=not-an-iterable
+            if hasattr(block.block, "to_table_of_contents_items"):
+                items += block.block.to_table_of_contents_items(block.value)
+
+        if self.contact_details_id:
+            items += [{"url": "#contact-details", "text": "Contact details"}]
```
zerolab, line 80: @mr-nick17 this is the approach to add various elements to the TOC

the `table_of_contents` template tag (`{{ table_of_contents(page, "body", {"title": "Table of contents"}) }}`) currently only picks the up from a StreamField, rather than anything else

zerolab (changes requested):
This LGMT as it is.
There is one outstanding item - the ONS Embed block https://github.com/ONSdigital/dis-wagtail-alpha/blob/main/ons_alpha/jinja2/templates/components/streamfield/ons_embed_block.html needs a bit of magic as currently it outputs a plain iframe

mr-nick17:
> This LGMT as it is. There is one outstanding item - the ONS Embed block https://github.com/ONSdigital/dis-wagtail-alpha/blob/main/ons_alpha/jinja2/templates/components/streamfield/ons_embed_block.html needs a bit of magic as currently it outputs a plain iframe

I could get the content to load in the iframe; I used (https://www.ons.gov.uk/visualisations/dvc2975/dashwrapper/index.html). But I needed to manually adjust the height which is automatically set by this [(dirty) jQuery](https://github.com/ONSdigital/babbage/blob/6088ffb0b80073604fefa16a24221f0d7006a11a/src/main/web/js/app/pym-interactive.js#L1). 

I will push an update that wraps it in a `<section>` and populates the title. What do you think re: height setting? Should we embed some vanilla JS that does the same? The overall look doesn't look quite right either, as the current iframe contents have the whole width of the useable grid but (my design) has it split to fit in the TOC

zerolab:
@mr-nick17 if it is not too much trouble, a bit of vanilla JS to ajust the height would go a long way, I think.

As for the width, we should flag this as a potential issue to explore with design / Joe N et al.

zerolab:
While looking at something unrelated, I noticed there's a richer her component - https://service-manual.ons.gov.uk/design-system/components/hero#example.example-hero-dark.nunjucks - and was wondering if this would be a better fit for the title + summary. Then again, this is one to discuss with the design team as I feel this may be one for the gap analysis

mr-nick17:
> While looking at something unrelated, I noticed there's a richer her component - https://service-manual.ons.gov.uk/design-system/components/hero#example.example-hero-dark.nunjucks - and was wondering if this would be a better fit for the title + summary. Then again, this is one to discuss with the design team as I feel this may be one for the gap analysis

Agree we probably need a steer from UI design as there are a couple of hero variants to consider. It might be that the solution is to have a setting within wagtail that allows the content editor to choose between different options or maybe it's fixed for a content type 🤷‍♂️

zerolab (approved):
As discussed. Let's get this in, then the JS for the iframe can come in a separate PR

MebinAbraham (approved):
LGTM.

While looking at this I noticed the table component is broken on first selection (existing). Once you select the table header option then it works. Small bug in code that we need to fix separately.

PR #17, opened 2024-07-10
Add support for cell merging and td class

zerolab:
This stems from a question from Al. The basic Wagtail TableBlock supports cell merging as well as alignment.

Note that https://service-manual.ons.gov.uk/design-system/components/table doesn't support col/row spans and this is a potential gap

Wagtail docs reference: https://docs.wagtail.org/en/stable/reference/contrib/table_block.html#configuration-options

### How to review
Add/edit an information page or a bulletin page, add a table block in content, right click on a cell to change alignment. Select several cells to merge (although that is currently not reflected in the markup because of the DS component)

### DIFF
--- a/ons_alpha/core/blocks/markup.py
+++ b/ons_alpha/core/blocks/markup.py
@@ -1,3 +1,4 @@
+from django.utils.safestring import mark_safe
 from django.utils.text import slugify
 from wagtail import blocks
 from wagtail.contrib.table_block.blocks import DEFAULT_TABLE_OPTIONS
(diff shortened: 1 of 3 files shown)


ons_alpha/core/blocks/markup.py
```diff
@@ -32,23 +33,75 @@ class Meta:
...
+                    classnames[(meta["row"], meta["col"])] = self._to_ons_classname(meta["className"])
+                if "hidden" in meta:
+                    hidden[(meta["row"], meta["col"])] = meta["hidden"]
+
+        if value.get("mergeCells"):
+            for merge in value["mergeCells"]:
+                spans[(merge["row"], merge["col"])] = {
+                    "rowspan": merge["rowspan"],
+                    "colspan": merge["colspan"],
+                }
```
zerolab: note: these are lifted from Wagtail core.
`hidden`/`span` are ready to plug in when the component would support them

MebinAbraham (commented):
As discussed, would be good to extend the macro for our needs so that this doesn't look like a broken feature. DS changes are likely to be in Beta. Otherwise LGTM.

zerolab:
@MebinAbraham b1d65e5 overrides the DS table macro. It doesn't cover all edge cases (like rowspan=2 in the first two rows, when the first row is designated as a header), but should be enough for POC

MebinAbraham (approved):

PR #20, opened 2024-07-11
Add bulletin series feature with routable pages and automated release determination

islas104:
**Description**:
This PR adds the necessary functionality to support a deterministic and joined-up bulletin release series. The following features are included:

- Added a `BulletinSeriesPage` to act as a container for each type of release.
- Implemented routable pages to provide `/latest` and `/previous-releases` endpoints.
- Automated the determination of the latest and previous releases based on the `release_date`.

<details>
<summary>previous issues</summary>

**Known Issues**:
Currently, there is an issue with the `wagtailmath` package where it tries to import `wagtail.core`, which no longer exists in recent versions of Wagtail. The error message is:

django.template.library.InvalidTemplateLibrary: Invalid template library specified. ImportError raised when trying to load 'wagtailmath.templatetags.wagtailmath': No module named 'wagtail.core'

**Steps to Reproduce**:
1. Run `python manage.py migrate` in the local development environment.

**Attempted Fixes**:
- Updated the import path in the `wagtailmath` package to use the new `wagtail` module:

```python
from wagtail.core.blocks import FieldBlock

to:

from wagtail.blocks import FieldBlock

```
Despite this change, the issue persists.

### Steps Taken:
- Attempted to update the import paths in the `wagtailmath` package to align with the new Wagtail version.
- Followed suggested fixes but the issue persists.

</details>

### Request for Feedback:
I am submitting this PR to gather feedback and suggestions from the team. If anyone has encountered similar issues or has ideas for a fix, your input would be greatly appreciated.

**Checklist**:
- [x] Add a bulletin series page to act as a container for each type of release.
- [x] Use `RoutablePageMixin` to provide `/latest` and `/previous-releases`.
- [x] Automate the sorting and determination of the latest and previous releases based on release date.
- [x] Add view latest release box/link to older bulletins.
- [x] Ensure the previous releases box is displayed on `/latest` and the latest bulletin page.

Additional Notes:
This PR is part of the POC and aims to demonstrate the core functionality. Further refinements and fixes will be made in subsequent iterations.

Screencast: https://www.loom.com/share/455a5c6d05ba48aea046a84c2e1a21bc?sid=ff166e0b-560e-44d3-bb07-c25f5d5628a1


ons_alpha/bulletins/blocks.py
```diff
@@ -13,13 +13,12 @@
...
 )
 
-
 class BulletinStoryBlock(StreamBlock):
     heading = HeadingBlock(show_back_to_toc=True)
     rich_text = RichTextBlock()
     panel = PanelBlock()
     table = TableBlock(group="DataVis")
-    equation = MathBlock(icon="decimal", group="DataVis")
+    equation = FieldBlock(icon="decimal", group="DataVis")  # Ensure the use of the correct FieldBlock
```
zerolab: wagtailmath is pinned to a commit that is compatible with Wagtail 5.2
See https://github.com/ONSdigital/dis-wagtail-alpha/blob/main/pyproject.toml#L24 / https://github.com/ONSdigital/dis-wagtail-alpha/blob/main/poetry.lock#L2091-L2113

Can you try:
```sh
fab sh-root
poetry install
```

or if you're running it directly on your machine, then in your active Python virtual environment, `pip install poetry` and then `poetry install`
?

ons_alpha/jinja2/templates/pages/bulletins/bulletin_page.html
```diff
@@ -1,17 +1,10 @@
-{% extends "templates/base_page.html" %}
+{% extends "base_page.html" %}
```
zerolab: @islas104 can you please rebase your branch off latest `main`?

the bulletin page template was updated in #14

ons_alpha/urls.py
```diff
@@ -14,6 +14,22 @@
 from ons_alpha.search import views as search_views
 from ons_alpha.utils.cache import get_default_cache_control_decorator
 
+from wagtail.contrib.routable_page.models import RoutablePageMixin, route
+from ons_alpha.bulletins.views import latest_bulletin_view, previous_releases_view
+
+class BulletinSeriesPage(RoutablePageMixin, Page):
```
zerolab: This is a page model. Models should live in their app's `model.py`. 

In this case, `ons_alpha/bulletins/models.py`
There is no need to separate latest/previous in a view module as the logic here is fairly simple.

Adding a new model should require Django migrations (so, `./manage.py makemigrations` in the container or if you're running it directly, then in your active venv)

ons_alpha/urls.py
```diff
@@ -14,6 +14,22 @@
...
+    parent_page_types = ['home.HomePage']
+    subpage_types = ['BulletinPage']
+
+    @route(r'^latest/$')
+    def latest_bulletin(self, request):
+        return latest_bulletin_view(request, self)
+
+    @route(r'^previous-releases/$')
+    def previous_releases(self, request):
+        return previous_releases_view(request, self)
```
zerolab: both these methods do `page.get_latest_release()` / `page.get_previous_releases()`. You defined those methods under `BulletinPage`, but here you are passing an instance of `BulletinSeriesPage` which will explode

ons_alpha/jinja2/templates/pages/bulletins/bulletin_page.html
```diff
@@ -1,17 +1,10 @@
...
 
 {% block main %}
     {{ table_of_contents(page, "body", {"title": "Table of contents"}) }}
-
     {% include_block page.updates %}
-
     {% include_block page.body %}
-
+    <div class="previous-releases">
+        <a href="{% pageurl page.get_parent.specific %}previous-releases/">Previous Releases</a>
```
zerolab: 1. you don't need specific for generating URLs. This is all encapsulated by the core `Page` class in Wagtail
2. I am fairly certain Jinja requires methods to be called as such (so `page.get_parent()`)

```suggestion
        <a href="{% pageurl page.get_parent() %}previous-releases/">Previous Releases</a>
```

ons_alpha/bulletins/models.py
```diff
@@ -58,3 +60,13 @@ class BulletinPage(BasePage):
...
     )
+
+    def get_latest_release(self):
+        parent = self.get_parent().specific
+        latest = BulletinPage.objects.live().descendant_of(parent).order_by('-release_date').first()
+        return latest
+
+    def get_previous_releases(self):
+        parent = self.get_parent().specific
+        previous = BulletinPage.objects.live().descendant_of(parent).exclude(id=self.id).order_by('-release_date')
```
zerolab: Better to use child of:

```suggestion
        parent = self.get_parent()
        previous = BulletinPage.objects.live().child_of(parent).exclude(id=self.id).order_by('-release_date')
```

ref: https://docs.wagtail.org/en/stable/reference/pages/queryset_reference.html#wagtail.query.PageQuerySet.child_of

zerolab: however, this should move to `BulletinSeriesPage` so this would become:

```suggestion
        previous = BulletinPage.objects.live().child_of(self).order_by('-release_date')
```

ons_alpha/bulletins/models.py
```diff
@@ -58,3 +60,13 @@ class BulletinPage(BasePage):
             ObjectList(BasePage.settings_panels, heading="Settings", classname="settings"),
         ]
     )
+
+    def get_latest_release(self):
+        parent = self.get_parent().specific
+        latest = BulletinPage.objects.live().descendant_of(parent).order_by('-release_date').first()
```
zerolab:
```suggestion
        parent = self.get_parent()
        latest = BulletinPage.objects.live().child_of(parent).order_by('-release_date').first()
```

zerolab: however, this should move to `BulletinSeriesPage` so this would become:

```suggestion
        latest = BulletinPage.objects.live().child_of(self).order_by('-release_date').first()
```

ons_alpha/jinja2/templates/pages/bulletins/bulletin_page.html
```diff
@@ -1,17 +1,10 @@
...
+{% extends "base_page.html" %}
 
 {% block main %}
     {{ table_of_contents(page, "body", {"title": "Table of contents"}) }}
-
     {% include_block page.updates %}
-
     {% include_block page.body %}
-
+    <div class="previous-releases">
```
zerolab: This doesn't take into account whether it this bulletin page is the latest, or not.

I suggest adding an `is_latest` property on `BulletinPage` that does the following check:

```python
class BulletinPage(...):
    ...

    @property
    def is_latest(self):
        latest_id = ReleasePage.objects.sibling_of(self).live().order_by('-release_date').values_list('id', flat=True).first()
         return self.pk == latest_id
```

ref: https://docs.wagtail.org/en/stable/reference/pages/queryset_reference.html#wagtail.query.PageQuerySet.sibling_of

requirements.txt
```diff
@@ -0,0 +1,42 @@
+# Django and Wagtail
```
zerolab: we use [poetry](https://python-poetry.org/docs/), so this requirements file is extraneous

requirements.txt
```diff
@@ -0,0 +1,42 @@
...
+telepath>=0.3.1,<1
+laces>=0.1,<0.2
+
+# Additional dependencies
+django-crispy-forms>=1.14.0,<2
+tbxforms>=1.0.3,<2
+whitenoise>=6.0,<7.0
+django-extensions>=3.1,<4.0
+django_jinja>=2.10,<3.0
+wagtailmath>=1.0,<2
```
zerolab: and this would explain your issues with wagtailmath

zerolab (changes requested):
Hey @islas104,

this is a good start. It needs more work, including a rebase on the `main` branch

ons_alpha/jinja2/templates/pages/bulletins/views.py
```diff
@@ -0,0 +1,10 @@
+from django.shortcuts import render
```
zerolab: This view is in the templates directory, but urls.py uses `from ons_alpha.bulletins.views import latest_bulletin_view, previous_releases_view`

which causes Django to fail to start all together

ons_alpha/jinja2/templates/pages/bulletins/bulletin_page.html
zerolab: this is also missing the "view latest" box for older bulletins

islas104:
Hi @zerolab ,

I've made all the requested changes and updated the code accordingly. The linting and other tests have passed. However, I'm running into an issue with the SECRET_KEY setting during the check. The error message indicates that the SECRET_KEY must not be empty.

As a workaround, I'm considering setting a default SECRET_KEY directly in the settings file for the purpose of these checks. Do you have a better suggestion or a preferred method to handle this?

zerolab:
@islas104 you can copy ons_alpha/settings/local.py.example to ons_alpha/settings/local.py and set it there

ons_alpha/bulletins/blocks.py
zerolab: @islas104 this is a no go. Please revert to MathBlock, as it was

ons_alpha/bulletins/models.py
```diff
@@ -61,22 +57,25 @@ class BulletinPage(BasePage):
...
 
-        if self.contact_details_id:
-            items += [{"url": "#contact-details", "text": "Contact details"}]
+    @route(r"^latest/$")
+    def latest_bulletin(self, request):
+        latest = BulletinPage.objects.live().child_of(self).order_by("-release_date").first()
+        return render(request, "templates/pages/bulletins/bulletin_page.html", {"page": latest})
 
-        return items
+    @route(r"^previous-releases/$")
```
zerolab:
```suggestion
    @path("previous-releases/")
```

ons_alpha/bulletins/models.py
```diff
@@ -61,22 +57,25 @@ class BulletinPage(BasePage):
...
-            if hasattr(block.block, "to_table_of_contents_items"):
-                items += block.block.to_table_of_contents_items(block.value)
+class BulletinSeriesPage(RoutablePageMixin, Page):
+    template = "templates/pages/bulletins/series_page.html"
+    parent_page_types = ["home.HomePage"]
+    subpage_types = ["BulletinPage"]
 
-        if self.contact_details_id:
-            items += [{"url": "#contact-details", "text": "Contact details"}]
+    @route(r"^latest/$")
```
zerolab: best to use `wagtail.contrib.routable_page.models.path`

```suggestion
    @route("latest/")
```

as you don't need any regex-ing.

The import will then have to change from `from wagtail.contrib.routable_page.models import RoutablePageMixin, route` to `from wagtail.contrib.routable_page.models import RoutablePageMixin, path`

ons_alpha/jinja2/templates/pages/bulletins/bulletin_page.html
zerolab: It is good to see a rebase, but now this change throws away Nicks work. Can you please tweak this so you only add your changes, rather than undo the template work

ons_alpha/jinja2/templates/pages/bulletins/bulletin_page.html
zerolab: You added the property on the model, and now it needs handling in the template

ons_alpha/jinja2/templates/pages/bulletins/bulletin_page.html
```diff
@@ -1,85 +1,10 @@
...
-                    <span class="ons-u-nowrap">{{ page.next_release_date|date("j F Y") }}</span>
-                </div>
-            {% endif %}
-
-        {% endblock release_meta %}
+    {{ table_of_contents(page, "body", {"title": "Table of contents"}) }}
+    {% include_block page.updates %}
+    {% include_block page.body %}
+    <div class="previous-releases">
+        <a href="{% pageurl page.get_parent() %}previous-releases/">Previous Releases</a>
```
zerolab:
```suggestion
        <a href="{% routablepageurl page.get_parent() "previous_release" %}">Previous Releases</a>
```

as per https://docs.wagtail.org/en/stable/reference/contrib/routablepage.html#the-routablepageurl-template-tag

ons_alpha/bulletins/models.py
```diff
@@ -61,22 +57,25 @@ class BulletinPage(BasePage):
...
+        return self.pk == latest_id
 
-    @cached_property
-    def toc(self):
-        items = [{"url": "#summary", "text": "Summary"}]
 
-        for block in self.body:  # pylint: disable=not-an-iterable
-            if hasattr(block.block, "to_table_of_contents_items"):
-                items += block.block.to_table_of_contents_items(block.value)
+class BulletinSeriesPage(RoutablePageMixin, Page):
```
zerolab, line 102: Now that the new page model is in the correct place, you need to make a new migration as per my previous comment.

The check for missing migrations is failing - 
https://github.com/ONSdigital/dis-wagtail-alpha/actions/runs/9896847664/job/27339953666?pr=20:

```
Run poetry run ./manage.py makemigrations --check --noinput
Skipping virtualenv creation, as specified in config file.
Migrations for 'bulletins':
  ons_alpha/bulletins/migrations/0003_bulletinseriespage_alter_bulletinpage_body_and_more.py
    - Create model BulletinSeriesPage
    - Alter field body on bulletinpage
    - Alter field updates on bulletinpage
```

You need to run `python manage.py makemigrations` (but after you undo the MathJax change), please.

zerolab (changes requested):

islas104:
Hi @zerolab ,

I've implemented all the changes you requested, including updating the routes and template adjustments. However, I'm encountering an error related to wagtail.core.blocks, which seems to be causing issues during the migration process.

The error message I'm seeing is:

ModuleNotFoundError: No module named 'wagtail.core'

I suspect this might be due to changes in the Wagtail package structure. Could you advise on the best way to address this? Should we update the imports to use wagtail.blocks or is there another preferred method to resolve this issue?

zerolab:
> The error message I'm seeing is:
> ModuleNotFoundError: No module named 'wagtail.core'
> 
> I suspect this might be due to changes in the Wagtail package structure. Could you advise on the best way to address this? Should we update the imports to use wagtail.blocks or is there another preferred method to resolve this issue?

`poetry install` please

ons_alpha/bulletins/models.py
```diff
@@ -61,22 +57,25 @@ class BulletinPage(BasePage):
         ]
     )
 
-    def get_context(self, request, *args, **kwargs):
```
zerolab, line 64: please don't remove functional code, unless there is an explicit need. And if so, do add a comment on the PR as to why it is removed.

This and the toc property are important and come from #14

ons_alpha/jinja2/templates/pages/bulletins/bulletin_page.html
```diff
@@ -1,6 +1,10 @@
 {% extends "templates/base_page.html" %}
 
 {% from "components/table-of-contents/_macro.njk" import onsTableOfContents %}
+{% block content %}
+    {% include_block page.updates %}
+    {% include_block page.body %}
+{% endblock %}
```
zerolab: These are extrneous in the template now
```suggestion
```

ons_alpha/jinja2/templates/pages/bulletins/bulletin_page.html
```diff
@@ -46,7 +49,7 @@ <h1 class="ons-u-fs-xxxl ons-u-mb-m">
             {# fmt:on #}
         </div>
 
-        <div class="ons-grid__col ons-col-8@m ons-col-12@s ons-u-p-no@xxs@m">
+        <div class="ons-grid__col ons-col-8@m ons-col-12@s ons-col-12@s ons-u-p-no@xxs@m">
```
zerolab: unnecessary change

ons_alpha/jinja2/templates/pages/bulletins/bulletin_page.html
```diff
@@ -77,6 +80,9 @@ <h3 class="ons-u-mb-no ons-u-fs-r--b">Phone</h3>
         </div>
     </div>
 
+    <div class="previous-releases">
```
zerolab: this has now lost the "view latest"

ons_alpha/settings/base.py
```diff
@@ -52,6 +52,9 @@
 if "CSRF_TRUSTED_ORIGINS" in env:
     CSRF_TRUSTED_ORIGINS = env["CSRF_TRUSTED_ORIGINS"].split(",")
 
+WAGTAILADMIN_BASE_URL = env.get("WAGTAILADMIN_BASE_URL", "http://localhost:8000")
```
zerolab: already defined in https://github.com/ONSdigital/dis-wagtail-alpha/blob/main/ons_alpha/settings/base.py#L691-L695

if you need to make changes to any of the already defined settings locally, please add them to `ons_alpha/settings/local.py`

ons_alpha/settings/base.py
```diff
@@ -128,6 +131,7 @@
         "BACKEND": "django_jinja.jinja2.Jinja2",
         "DIRS": [
             PROJECT_DIR / "jinja2",
+            PROJECT_DIR / "templates",  # Ensure this is included
```
zerolab: these changes should not be necessary

ons_alpha/jinja2/templates/pages/bulletins/previous_releases.html
```diff
@@ -0,0 +1,10 @@
+{% extends "base_page.html" %}
```
zerolab:
```suggestion
{% extends "templates/base_page.html" %}
```

ons_alpha/bulletins/models.py
```diff
@@ -61,22 +59,43 @@ class BulletinPage(BasePage):
...
+        latest = BulletinPage.objects.live().child_of(self).order_by("-release_date").first()
+        return render(request, "templates/pages/bulletins/bulletin_page.html", {"page": latest})
+
+    @path("previous-releases/")
+    def previous_releases(self, request):
+        previous = BulletinPage.objects.live().child_of(self).order_by("-release_date")
+        return render(request, "templates/pages/bulletins/previous_releases.html", {"bulletins": previous})
+
+    def serve(self, request, *args, **kwargs):
+        return redirect(self.url + self.reverse_subpage("latest_bulletin"))
```
RealOrangeOne: Nit: It's bad practice to use `.url` directly. By using `.get_url(request)`, Wagtail can cache some things on the request, and getting the URL is faster (as it doesn't need to do any DB queries).

ons_alpha/bulletins/models.py
```diff
@@ -61,22 +59,43 @@ class BulletinPage(BasePage):
...
+
+
+class BulletinSeriesPage(RoutablePageMixin, Page):
+    template = "templates/pages/bulletins/series_page.html"
+    parent_page_types = ["home.HomePage"]
+    subpage_types = ["BulletinPage"]
+
+    @path("latest/")
+    def latest_bulletin(self, request):
+        latest = BulletinPage.objects.live().child_of(self).order_by("-release_date").first()
```
RealOrangeOne: Suggestion: Might be nice to move this to a separate method on the `BulletinSeriesPage`, so it can be reused here and below, and perhaps in the `is_latest` property above, too.

ons_alpha/bulletins/models.py
```diff
@@ -61,22 +59,43 @@ class BulletinPage(BasePage):
...
+        return render(request, "templates/pages/bulletins/bulletin_page.html", {"page": latest})
```
RealOrangeOne: Question: Should this render the page directly, or just redirect to it instead?

zerolab: render directly. `/latest` is an evergreen URL - https://www.ons.gov.uk/economy/inflationandpriceindices/bulletins/consumerpriceinflation/latest

islas104:
@zerolab - are you able to provide feedback for these changes?

zerolab:
Tidied the PR up and recorded a screencast: https://www.loom.com/share/455a5c6d05ba48aea046a84c2e1a21bc?sid=ff166e0b-560e-44d3-bb07-c25f5d5628a1

@MebinAbraham can you give your seal of approval so I merge this?

MebinAbraham (approved):

zerolab (approved):

PR #21, opened 2024-07-12
Feature/add custom iframe js

mr-nick17:
Adding enhanced functionality for iframes. The current implementation uses some custom JS and the `pym` library to resize the iframe height to fit the height of the content within.

Notes 
- I have removed the ability for the user to embed the iframed content (as currently). This requires further discussion/requirements before implementing.
- I've (just) noticed that page does cache the content before the JS has loaded and finished resizing; do a hard refresh to see the frame fully load

### How to review
Sense check
Image check

image


zerolab (approved):
LGTM!

The main content width is an issue we will want to highlight, but within the DS constraints this works nicely

mr-nick17:
> LGTM!
> 
> The main content width is an issue we will want to highlight, but within the DS constraints this works nicely

Yep the width is definitely a bit of an eyesore. There's also the bit about including 'embed this code' functionality to discuss with UCD too

PR #23, opened 2024-07-16
Add taxonomy management

MebinAbraham:
Add support for creating Topics/sub-topics via the admin UI and linking these topics to Topic Section, Topic Page and Bulletin pages.

[Demo](https://www.loom.com/share/22856b71f5344ea4b85ae6376f480426?sid=50c3eb3c-7834-42f5-9ee9-53382a9b740f)

Details:
- Topics are manually linked to Topic Pages. A topic/sub topic can only be linked to one topic/topic section page. It is a 1:1 relationship.
- Topic section page will only display top-level topics
- Topic page (sub-topic) will only show subtopics.
- Bulletins can link to zero or more topics at any level.
- Not all edge cases have been covered such as moving child to root etc.

### How to review
- Ensure the change makes sense
- Not missing any implementation

Known issue:
- Currently, when a topic is linked to a Topic/Topic Section page using the One-to-One relationship, the references do not seem to populate correctly when deleting the Topic itself. When using a ForeignKey, the reference is calculated correctly. This needs further investigation.


ons_alpha/bulletins/models.py
```diff
@@ -53,6 +62,13 @@ class BulletinPage(BasePage):
     edit_handler = TabbedInterface(
         [
             ObjectList(content_panels, heading="Content"),
+            ObjectList(
+                [
+                    MultipleChooserPanel("topics", label="Topic", chooser_field_name="topic"),
```
zerolab, line 74: 👌🏼 for the `MultipleChooserPanel` use

ons_alpha/taxonomy/admin.py
```diff
@@ -0,0 +1 @@
+# Register your models here.
```
zerolab: nitpick: not needed, unless you plan to customise things for Django admin

ons_alpha/taxonomy/models.py
```diff
@@ -0,0 +1,101 @@
...
+from treebeard.mp_tree import MP_Node
+from wagtail.admin.forms import WagtailAdminModelForm
+from wagtail.admin.panels import FieldPanel
+from wagtail.search import index
+
+
+# This is the main 'node' model, it inherits mp_node
+# mp_node is short for materialized path, it means the tree has a clear path
+class Topic(index.Indexed, MP_Node):
+    name = models.CharField(max_length=30)
```
zerolab: Note that 30 is perhaps too short. e.g. https://www.ons.gov.uk/economy/governmentpublicsectorandtaxes ("Government, public sector and taxes") won't fit

ons_alpha/taxonomy/models.py
```diff
@@ -0,0 +1,101 @@
...
+
+    search_fields = [
+        index.FilterField("name"),
+        index.FilterField("depth"),
+        index.SearchField("name"),
+        index.AutocompleteField("name"),
+    ]
+
+    class Meta:
+        verbose_name_plural = "Topics"
```
zerolab: note: this should not be needed as Django automatically appends `s` to the end of the model name. This is useful if you need to tweak things, or the model ends in "s"

ons_alpha/taxonomy/views.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class TopicChooserViewSet(ChooserViewSet):
+    model = Topic
+
+    icon = "tag"
+    choose_one_text = "Choose a topic"
+    choose_another_text = "Choose another topic"
+    url_filter_parameters = ["country"]
+    preserve_url_parameters = ["multiple", "country"]
```
zerolab: these won't do anything as we don't have a country field.
This looks like a remnant from https://docs.wagtail.org/en/stable/extending/generic_views.html#limiting-choices-via-linked-fields

It should be possible to use the `linked_fields` param on the widget with a regex, on the panel definition on the Bulletin page, will check tomorrow

ons_alpha/taxonomy/views.py
```diff
@@ -0,0 +1,63 @@
...
+class TopicChooserMixin:
+    def get_object_list(self):
+        # Only show top-level topics
+        return Topic.get_root_nodes()
+
+
+class SubTopicChooserMixin:
+    def get_object_list(self):
+        # Only show topic that have a parent
+        return Topic.objects.exclude(depth=1).order_by("path")
```
zerolab: I am not sure we need the extra complexity, especially that we can have more than 2 levels. In the screenshot I have a topic section under another topic section, 
Screenshot 2024-07-16 at 14 32 00

but I only get the top level in the chooser:

Screenshot 2024-07-16 at 14 32 08

ons_alpha/topics/models.py
```diff
@@ -1,15 +1,81 @@
...
+        Topic,
+        on_delete=models.SET_NULL,
+        null=True,
+        blank=False,
+        related_name="topic_page",
+    )
+
+    content_panels = BasePage.content_panels
+
+    edit_handler = TabbedInterface(
```
zerolab, line 23: nice use of the tabbed inderface.

I'd suggest setting the widget class directly in the field panel definition, rather than via `_taxonomy_widget_class` thought. We don't need it outside of that. Also, if we keep to the one chooser, then you don't need to specify the widget at all

ons_alpha/bulletins/models.py
```diff
@@ -141,3 +157,8 @@ def previous_releases(self, request):
             context_overrides={"bulletins": previous},
             template="templates/pages/bulletins/previous_releases.html",
         )
+
+
+class BulletinTopicRelationship(Orderable):
```
zerolab: note on the convention: we tend to define the orderable above its related page model. something, something locality of behaviour and we've always done it.

not a blocker

zerolab (changes requested):
Did a first pass check. Will continue tomorrow morning

ons_alpha/taxonomy/admin.py
MebinAbraham: Removed

ons_alpha/taxonomy/models.py
MebinAbraham: Good shout, updated to 100 for safe measure.

ons_alpha/taxonomy/models.py
MebinAbraham: Forgot about this, removed.

ons_alpha/taxonomy/views.py
MebinAbraham: Whoops, yup just remnants from testing. Removed.

ons_alpha/taxonomy/views.py
MebinAbraham: I was thinking even "sub TopicSections" would be top-level taxonomy Topics. But for now, I've removed the filtering and kept it simple.

ons_alpha/topics/models.py
MebinAbraham: Only a single chooser now so removed the need to pass widget class.

ons_alpha/bulletins/models.py
MebinAbraham: Good to know. Updated.

ons_alpha/taxonomy/models.py
```diff
@@ -0,0 +1,98 @@
...
+        index.SearchField("name"),
+        index.AutocompleteField("name"),
+    ]
+
+    def __str__(self):
+        return self.name_with_depth()
+
+    # this is just a convenience function to make the names appear with lines
+    # eg root | - first child
+    def name_with_depth(self):
```
zerolab, line 36: Side note: Wagtail's Collection object has a method that does the indentation - https://github.com/wagtail/wagtail/blob/2d568dd825e4b087d7cea1c4d1fa49d916b395c2/wagtail/models/media.py#L93 - one we could use in the future.

This works fine for our needs

zerolab (approved):

PR #25, opened 2024-07-18
Custom scheduler

RealOrangeOne:
https://jira.ons.gov.uk/browse/NWP-262

It's critical that pages are published as close to their publish time as possible. Using an external scheduler adds additional latency. Instead, run a permanent process which handles scheduling of jobs. This ensures the process is already running, and can immediately execute jobs. 

From testing, jobs are executed almost immediately (single-digit milliseconds). A simple run of `publish_scheduled` locally with a single page to publish took 50ms (this **will** get larger as the site content grows).

For `honcho` to work, you may need to completely rebuild the container (and/or recreate the virtualenv).

### How to review

This PR adds a new management command `scheduler` which runs the relevant commands.

### DIFF
--- a/heroku.yml
+++ b/heroku.yml
@@ -7,3 +7,8 @@ release:
   image: web
   command:
     - django-admin createcachetable && django-admin migrate --noinput
+run:
+  scheduler:
+    image: web
+    command:
+      - django-admin scheduler
--- a/ons_alpha/core/management/commands/scheduler.py
+++ b/ons_alpha/core/management/commands/scheduler.py
@@ -1,0 +1,35 @@
+import atexit
+import logging
+import signal
+
+from apscheduler.executors.pool import ThreadPoolExecutor
+from apscheduler.schedulers.blocking import BlockingScheduler
+from django.core.management.base import BaseCommand
+
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    def handle(self, **options):
+        self.scheduler = BlockingScheduler(executors={"default": ThreadPoolExecutor()})
+
+        self.setup_signals()
+
+        self.configure_scheduler()
+
+        logger.info("Starting scheduler")
+        self.scheduler.start()
+
+    def setup_signals(self):
+        signal.signal(signal.SIGINT, self.shutdown)
+        signal.signal(signal.SIGTERM, self.shutdown)
+        atexit.register(self.shutdown)
+
+    def shutdown(self, *args, **kwargs):
+        if self.scheduler.running:
+            logger.info("Shutting down scheduler")
+            self.scheduler.shutdown(wait=False)
+
+    def configure_scheduler(self):
+        pass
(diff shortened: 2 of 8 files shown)


ons_alpha/core/management/commands/scheduler.py
```diff
@@ -0,0 +1,37 @@
...
+    def shutdown(self, _signum, _frame):
+        if self.scheduler.running:
+            self.scheduler.shutdown(wait=False)
+
+    def add_management_command(self, command_name, trigger, **kwargs):
+        func = partial(call_command, command_name, **kwargs)
+        self.scheduler.add_job(func, name=command_name, trigger=trigger)
+
+    def configure_scheduler(self):
+        self.add_management_command("publish_scheduled", CronTrigger(second=0))
```
zerolab, line 38: so this runs every minute on the minute, right?
suggestion: add a dev note

pyproject.toml
```diff
@@ -123,7 +125,3 @@ omit = [
 
 [tool.coverage.report]
 show_missing = true
-
-[tool.poetry.plugins.dotenv]
```
zerolab, line 127: dev note: this caused issues with honcho

heroku.yml
```diff
@@ -7,3 +7,8 @@ release:
   image: web
   command:
     - django-admin createcachetable && django-admin migrate --noinput
+run:
+  scheduler:
+    image: web
+    command:
+      - django-admin scheduler
```
zerolab, line 14: reminder: we'll need a separate task for the ONS infra, when it comes

zerolab (approved):
Left a suggestion and reminder.

I had to manually remove `/venv/lib/python3.12/site-packages/*`, then rebuild the container for honcho to start working

ons_alpha/core/management/commands/scheduler.py
RealOrangeOne: This caught me out too, but yes. It runs when when the `second` value equals `0` (ie every minute on the minute, much like cron).

heroku.yml
RealOrangeOne: We should confirm whether the Kubernetes `CronJob` is suitable. I've created a ticket to track benchmarking it: https://jira.ons.gov.uk/browse/NWP-274

MebinAbraham (approved):
This needs documenting in our confluence space. We need to cover why we are trying this, what are the benefits and drawbacks and what other options are we going to explore.

Can you update https://confluence.ons.gov.uk/display/NWP/Wagtail+Scheduling

RealOrangeOne:
I've populated the page, and mentioned that the scheduler is merely a stop-gap until we've done some more benchmarking.

MebinAbraham:
> I've populated the page, and mentioned that the scheduler is merely a stop-gap until we've done some more benchmarking.

Thanks Jake. Looks like a good starting point until we do further investigation. 👍

PR #26, opened 2024-07-18
Send a notification to Slack on page publish

RealOrangeOne:

When a page is published, send a notification to Slack.

This is run on **every** page publish, as opposed to only ones of a certain type. Once collections are in, the hook can be modified.

Currently, the Slack webhook is a blocking operation.

### How to review

image

The webhook URL should be placed in `$SLACK_PUBLISH_NOTIFICATION_WEBHOOK_URL`.


ons_alpha/core/wagtail_hooks.py
```diff
@@ -13,3 +20,33 @@ def register_icons(icons):
...
     ]
+
+
+@hooks.register("after_publish_page", order=float("inf"))
+def notify_slack_of_publish(request, page):
+    if (webhook_url := settings.SLACK_PUBLISH_NOTIFICATION_WEBHOOK_URL) is None:
+        return
+
+    revision = page.latest_revision
+    user = revision.user
```
MebinAbraham, line 32: It would be good to also output the release date time if the page has that field, so it can be sanity-checked quickly.

ons_alpha/core/wagtail_hooks.py
```diff
@@ -13,3 +20,33 @@ def register_icons(icons):
...
+
+    client = WebhookClient(webhook_url)
+
+    response = client.send(
+        text="A new page has been published",
```
MebinAbraham: Minor, drop "new" as this applies to already published pages as well. Non-blocker but bonus points if you can update to say `republished` when already live. 🙂

MebinAbraham (commented):
Notification working as expected, just a couple of suggestions.

ons_alpha/core/wagtail_hooks.py
```diff
@@ -13,3 +21,40 @@ def register_icons(icons):
...
+        text="A new page has been published" if is_new_page else "A page has been updated",
+        attachments=[
+            {
+                "color": "good",
+                "fields": [
+                    {"title": "Title", "value": page.title, "short": True},
+                    {"title": "Published by", "value": user.get_full_name() if user else "System", "short": True},
+                    {
+                        "title": "Go-live time",
+                        "value": date_format(go_live_time, settings.SHORT_DATETIME_FORMAT),
```
MebinAbraham, line 48: nit: Can we update `SHORT_DATETIME_FORMAT`, it is currently American format.

MebinAbraham (approved):

PR #28, opened 2024-07-22
Bundling and bundled publishing

zerolab:
https://jira.ons.gov.uk/browse/NWP-260

- Adds a new bundles app, and model
- chooser viewsets, various buttons to add to bundle
- a mixin to use on pages to be bundled
- a management command dealing with publishing

To-Do
- [x] demo screencast
- [x] populate release calendar
- [x] some aditional validation (e.g. not allowing a release calendar that is already in an active bundle)
- [x] update the task scheduler to use the `publish_bundles` management command

Screencast: https://www.loom.com/share/9ec78d1cff8946b483339ff903ee0c1e?sid=ef77be80-ecd9-4c7a-8f56-d02e58e40a40


ons_alpha/bundles/models.py
```diff
@@ -0,0 +1,180 @@
...
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="bundles",
+    )
+    status = models.CharField(choices=BundleStatus.choices, default=BundleStatus.PENDING, max_length=32)
+
+    objects = BundleManager()
+    active_objects = ActiveBundlesManager()
+    editable_objects = EditableBundlesManager()
+
```
RealOrangeOne: Issue: I'm not a huge fan of this pattern, as it can get a little confusing to manage. Queryset methods for `.active` and `.editable` feels more Django-native.

ons_alpha/bundles/management/commands/publish_bundles.py
```diff
@@ -0,0 +1,100 @@
...
+                        # note: description here is a simple CharBlock
+                        "description": strip_tags(related.description),
+                        "external_url": related.url,
+                    },
+                }
+            )
+        if links:
+            content.append({"type": "release_content", "value": {"title": "Datasets", "links": links}})
+        page = bundle.release_calendar_page
+        page.content = json.dumps(content, cls=DjangoJSONEncoder)
```
RealOrangeOne: Nit: I think Wagtail will handle serializing this itself - no need to JSON encode it manually.

ons_alpha/bundles/management/commands/publish_bundles.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @transaction.atomic
+    def handle_bundle(self, bundle: Bundle):
+        bundled_pages = bundle.get_bundled_pages()
+
+        bundled_revisions = [page.scheduled_revision for page in bundled_pages if page.scheduled_revision]
+        for revision in bundled_revisions:
+            # just run publish for the revision -- since the approved go
+            # live datetime is before now it will make the object live
+            revision.publish(log_action="wagtail.publish.scheduled")
```
RealOrangeOne:
```suggestion
        for page in bundle.get_bundled_pages():
            if (revision := page.scheduled_revision) is None:
                continue
            # just run publish for the revision -- since the approved go
            # live datetime is before now it will make the object live
            revision.publish(log_action="wagtail.publish.scheduled")
```

ons_alpha/bundles/management/commands/publish_scheduled_without_bundles.py
```diff
@@ -0,0 +1,103 @@
...
+    def handle(self, *args, **options):
+        dry_run = False
+        if options["dry-run"]:
+            self.stdout.write("Will do a dry run.")
+            dry_run = True
+
+        self._unpublish_expired(dry_run)
+        self._publish_scheduled_without_bundles(dry_run)
+
+    def _publish_scheduled_without_bundles(self, dry_run):
```
RealOrangeOne: Nit: Given we call this method second, I'd recommend defining it second. Reading the class and seeing a comment about step 2 can be a little jarring.

ons_alpha/bundles/migrations/0002_bundlepages_bundlelinks.py
```diff
@@ -0,0 +1,60 @@
+# Generated by Django 4.2.14 on 2024-07-19 14:33
```
RealOrangeOne: Nit: Squash migrations

ons_alpha/bundles/models.py
```diff
@@ -0,0 +1,180 @@
...
+        FieldPanel("description"),
+    ]
+
+
+class BundleManager(models.Manager):
+    def get_queryset(self):
+        queryset = super().get_queryset()
+        queryset = queryset.annotate(
+            release_date=Coalesce("publication_date", "release_calendar_page__release_date")
+        ).order_by(F("release_date").desc(nulls_last=True), "-pk")
```
RealOrangeOne: Question: Ordering by `pk` can be a little confusing. Perhaps this should be `name` instead?

ons_alpha/bundles/models.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def save(self, **kwargs):
+        super().save(**kwargs)
+
+        if self.status == BundleStatus.RELEASED:
+            return
+
+        if self.scheduled_publication_date and self.scheduled_publication_date >= now():
+            # Schedule publishing for related pages
+            for bundled_page in self.get_bundled_pages().specific():
```
RealOrangeOne: Suggestion: Should this be `specific(defer=True)` or `.defer_streamfields`?

ons_alpha/bundles/panels.py
```diff
@@ -0,0 +1,22 @@
...
+
+        def _content_for_instance(self, instance):
+            if not hasattr(instance, "bundles"):
+                return ""
+
+            if bundles := instance.bundles:
+                content = "This page is in the following bundle(s):<ul>"
+
+                for bundle in bundles:
+                    content += f"<li>{bundle.name} (Status: {bundle.get_status_display()})</li>"
```
RealOrangeOne: Issue: This is vulnerable to XSS - I'd suggest using Django's [`format_html_join`](https://docs.djangoproject.com/en/5.0/ref/utils/#django.utils.html.format_html_join)

ons_alpha/bundles/views.py
```diff
@@ -0,0 +1,81 @@
...
+from .models import BundledPageMixin, BundlePage
+
+
+if TYPE_CHECKING:
+    from django.http import HttpRequest
+
+
+def add_to_bundle(request: "HttpRequest", page_to_add_id: "Page"):
+    page_to_add = get_object_or_404(Page, id=page_to_add_id)
+    page_to_add = page_to_add.specific
```
RealOrangeOne: Suggestion: This should work, and save a line
```suggestion
    page_to_add = get_object_or_404(Page.objects.specific(), id=page_to_add_id)
```

ons_alpha/bundles/views.py
```diff
@@ -0,0 +1,81 @@
...
+
+    if not issubclass(type(page_to_add), BundledPageMixin):
+        raise PermissionDenied
```
RealOrangeOne: Nit: Should this be a 400 rather than 403?

ons_alpha/bundles/wagtail_hooks.py
```diff
@@ -0,0 +1,60 @@
...
+
+class PageAddToBundleButton(PageListingButton):
+    label = _("Add to Bundle")
+    icon_name = "boxes-stacked"
+    aria_label_format = _("Add '%(title)s' to a bundle")
+    url_name = "bundles:add_to_bundle"
+
+    @property
+    def show(self) -> bool:
+        if not issubclass(type(self.page), BundledPageMixin):
```
RealOrangeOne: Nit: `issubclass(type(T), C)` is the same as `isinstance(T, C)`, which might be a little clearer.

ons_alpha/core/management/commands/scheduler.py
```diff
@@ -35,4 +35,5 @@ def add_management_command(self, command_name, trigger, **kwargs):
 
     def configure_scheduler(self):
         # "second=0" run the task every minute, on the minute (ie when the seconds = 0)
-        self.add_management_command("publish_scheduled", CronTrigger(second=0))
+        self.add_management_command("publish_bundles", CronTrigger(second=0))
+        self.add_management_command("publish_scheduled_without_bundles", CronTrigger(minute=5))
```
RealOrangeOne: Question: Why are you only checking scheduled pages once an hour?

zerolab: gah, I wanted it to run every 5 min 🤦🏼 
will re-read the docs

ons_alpha/bundles/management/commands/publish_bundles.py
zerolab: `page.content = content` does the trick. This was the remnant of some trials (tl;dr - you need to set an id on the nested structs for them to be picked up properly)

ons_alpha/bundles/migrations/0002_bundlepages_bundlelinks.py
zerolab: oops, forgot! good shout

ons_alpha/bundles/views.py
zerolab: Will change to a 404 as it provides a better experience oob

ons_alpha/core/management/commands/scheduler.py
zerolab: Updated to `*/5` as per https://apscheduler.readthedocs.io/en/3.x/modules/triggers/cron.html#expression-types

ons_alpha/bundles/models.py
zerolab: Both are valid and have their place. I agree this is probably overkill for our use case, so I moved 'em to the queryset

RealOrangeOne (approved):

ons_alpha/bundles/models.py
zerolab: how about `name`, then `-pk`?

zerolab: went with the latter. should probably make name unique, but that is for later

PR #29, opened 2024-07-24
Track and serve corrections

RealOrangeOne:
Replaces #27 

This PR adds the ability to log "corrections" linked to a specific revision, which can then be exposed and served using a `/previous/` URL.

The biggest caveat with this is that the version number is implicitly the index of the correction block. This ought to be ok, but probably isn't ideal. We could additionally inject a version number into the block, so reordering / deleting entries doesn't reorder them.


ons_alpha/core/static/js/previous-version-block.js
```diff
@@ -0,0 +1,49 @@
...
+        placeholder.replaceWith(this.element);
+
+        this.renderContent(initialState);
+    }
+
+    renderContent(state) {
+        if (state) {
+            this.element.innerHTML = `
+                <a class="button button-small button-secondary" href="../revisions/${state}/view/">View revision</a>
+                <a class="button button-small button-secondary" href="../revisions/${state}/revert/">Review revision</a>
```
zerolab: suggestion: `targe="_blank"` for both as the context is lost

ons_alpha/bulletins/models.py
```diff
@@ -121,6 +121,25 @@ def get_context(self, request, *args, **kwargs):
...
 
+    @path("previous/v<int:version>/")
+    def previous_version(self, request, version):
+        if version <= 0:
+            raise Http404
+
+        corrections = self.updates.blocks_by_name("correction")
+
+        try:
+            correction = corrections[version - 1]
```
zerolab, line 138: thought: this currently hinges on the order of the blocks in the, rather than when they were added. so if I re-order the blocks, their "version" change. This is OK for the POC.

The other thought. At the moment, it is possible to remove a correction. This should not be possible in the real world.


Both notes may be good to add in a HelpPanel above the corrections and notices field panel

ons_alpha/core/static/js/previous-version-block.js
zerolab: suggestion: either 
a. drop "Review revision" as when go to is, the actions are destruction - in that they promote that revision as the latest 
b. replace the action to our own that saves the revision in place

ons_alpha/core/blocks/panels.py
```diff
@@ -22,12 +26,34 @@ class Meta:
...
+        template = "templates/components/streamfield/corrections_block.html"
+
+
+class PreviousVersionBlockAdapter(FieldBlockAdapter):
+    js_constructor = "ons_alpha.core.blocks.panels.PreviousVersionBlock"
+
+    @cached_property
+    def media(self):
+        structblock_media = super().media
+        return Media(js=structblock_media._js + ["js/previous-version-block.js"], css=structblock_media._css)
```
zerolab, line 49: You can achieve the same via a custom `form_template` and passing the extra context in `get_form_context` on `CorrectionBlock`. See https://docs.wagtail.org/en/stable/advanced_topics/customisation/streamfield_blocks.html#custom-editing-interfaces-for-structblock

ons_alpha/jinja2/templates/components/streamfield/corrections_notices_block.html
```diff
@@ -1,5 +1,21 @@
 <h2>{{ _("Corrections and notices") }}</h2>
```
zerolab: I wonder if we should wrap these in https://service-manual.ons.gov.uk/design-system/components/details

as well as only display the h2 heading if there either corrections or notices. A small thing, but looks good on the FE ;)

zerolab (changes requested):
It is looking pretty good. Added a number of notes and suggestions

ons_alpha/bulletins/models.py
RealOrangeOne: I'm not sure Wagtail lets you do append-only blocks, but yes it's worth documenting at least. If it's a problem, we can POC a proper solution.

ons_alpha/core/blocks/panels.py
RealOrangeOne: You can, but also you can't. As part of rendering the template, I specifically need the value, which isn't present when `form_template` is rendered, as all that is handled client-side. So sadly, this JS is needed :disappointed: 

If we just wanted to hide the `previous_version` field, this would work. But to show the buttons we need its current value.

zerolab: yeah, that is annoying. let's leave it as it is

ons_alpha/jinja2/templates/components/streamfield/corrections_block.html
```diff
@@ -1,5 +1,8 @@
-<h3>{{ _("Corrections") }}</h3>
+{% from "components/details/_macro.njk" import onsDetails %}
 
-{% for item in value %}
-    {{ item }}
-{% endfor %}
+
+{% call onsDetails({"id": "correction-v" + version_num|string, "title": value.when|date("DATETIME_FORMAT")}) %}
```
zerolab: I did mean wrapping the whole `ons_alpha/jinja2/templates/components/streamfield/corrections_notices_block.html` in a details block, not individual corrections. This is the pattern currently used

RealOrangeOne: Oh, yes that's much closer to how it is currently - not sure how I missed that!

zerolab (approved):
separate thought: we can prevent re-ordering by makeing corrections inlines

but good to go as is for now

RealOrangeOne:
[Screencast_20240725_143207.webm](https://github.com/user-attachments/assets/e4d393f7-fc0d-4018-a6e0-db21d7cf4c47)

PR #31, opened 2024-07-29
Allow embedding documents in content

RealOrangeOne:

Add the ability to embed documents in content, styled by the DS: https://service-manual.ons.gov.uk/design-system/components/document-list.

Not all the DS features are exposed, as they require varying levels of extra work. But the core functionality is there, and can be expanded as time goes on.

### How to review

image

### DIFF
--- a/ons_alpha/core/blocks/embeddable.py
+++ b/ons_alpha/core/blocks/embeddable.py
@@ -1,4 +1,6 @@
+from django.conf import settings
 from django.core.exceptions import ValidationError
+from django.template.defaultfilters import filesizeformat
 from wagtail import blocks
 from wagtail.blocks import StructBlockValidationError
 from wagtail.documents.blocks import DocumentChooserBlock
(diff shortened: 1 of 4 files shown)


ons_alpha/core/blocks/embeddable.py
```diff
@@ -14,15 +16,47 @@ class Meta:
...
     title = blocks.CharBlock(required=False)
+    description = blocks.RichTextBlock(features=settings.RICH_TEXT_BASIC, required=False)
 
     class Meta:
         icon = "doc-full-inverse"
+        value_class = DocumentBlockStructValue
         template = "templates/components/streamfield/document_block.html"
 
 
+class DocumentsBlock(blocks.StreamBlock):
```
RealOrangeOne, line 46: A `ListBlock` would also have worked here, but subclassing `StreamBlock` is a better-defined path.

ons_alpha/core/blocks/embeddable.py
```diff
@@ -14,15 +16,47 @@ class Meta:
         template = "templates/components/streamfield/image_block.html"
 
 
+class DocumentBlockStructValue(blocks.StructValue):
+    def as_macro_data(self):
+        return {
+            "thumbnail": True,
+            "url": self["document"].url,
+            "title": self["title"] or self["document"].filename,
```
MebinAbraham: Since document title is mandatory, fall back to that.
```suggestion
            "title": self["title"] or self["document"].title,
```

MebinAbraham (approved):

PR #32, opened 2024-07-30
Integrate with the Dataset API

MebinAbraham:
This PR adds the ability to look up datasets via the ONS Dataset API.

**Notes:**
- For Beta, we likely need a custom chooser implementation to allow a stepped selection process (i.e., Select Dataset name, then Edition, then Version).
- Queryish might not be ideal for Beta as it is. We may need to look at a custom implementation to reduce the number of requests to the API.
- For Beta, we need to consider caching.
- This implementation doesn't account for user permissions. We still need to figure out how to map Wagtail users/groups with the Dataset API.

### How to review
Ensure:
- You can add datasets via lookup on the Release calendar entry page and a bundle.
- When datasets are added to a bundle and tied to a release calendar entry, ensure that on publish, the release calendar page contents are updated correctly.
- The changes make sense.

Screencasts: [Part 1](https://www.loom.com/share/bf7b238235a940cf8ce636cb9088b3a6?sid=8b074ffb-1d93-4267-9f15-3f84a284c0fe), [Part 2](https://www.loom.com/share/cd22d5b78b2c47a183beedb149cb0f57?sid=c62dc9d3-f6e8-4cd4-b1d0-b4ab3bf0072d)


ons_alpha/core/blocks/related.py
```diff
@@ -18,11 +18,16 @@
 class LinkBlockStructValue(StructValue):
     @cached_property
     def link(self) -> dict | None:  # pylint: disable=inconsistent-return-statements
+        title, desc = self.get("title"), self.get("description")
```
RealOrangeOne: Nit: This is clearer split onto multiple lines
```suggestion
        title = self.get("title")
        desc = self.get("description")
```

ons_alpha/datasets/views.py
```diff
@@ -0,0 +1,108 @@
...
+        if search_query:
+            search_query_lower = search_query.strip().lower()
+            # This is brittle, as it doesn't use dynamically set search fields
+            objects = [
+                obj
+                for obj in objects
+                if search_query_lower in obj.title.lower()
+                or search_query_lower in obj.formatted_edition.lower()
+                or search_query_lower in obj.version.lower()
+            ]
```
RealOrangeOne, line 46: Question: Can you use `.filter` here? It'd make the code easier to read

ons_alpha/datasets/views.py
```diff
@@ -0,0 +1,108 @@
...
+            self.is_searching = True
+            self.search_query = search_query
+        return objects
+
+
+class CustomFilterForm(CustomSearchFilterMixin, BaseFilterForm): ...
+
+
+class BaseView(BaseChooseView):
```
RealOrangeOne: Nit: `BaseView` is quite a generic name. Perhaps `BaseONSDatasetChooseView`?

ons_alpha/datasets/views.py
```diff
@@ -0,0 +1,108 @@
...
+        item = ONSDataset.objects.get(pk=pk)
+        pk = f"{item.id}__{item.edition}_{item.version}"
+        dataset, _ = Dataset.objects.get_or_create(
+            pk=pk,
+            title=item.title,
+            description=item.description,
+            version=item.version,
+            url=item.url,
+            edition=item.edition,
+        )
```
RealOrangeOne: Question: Do these attributes on a dataset (the remote one) change at all? If they do, this could result in duplicate `Dataset` instances.

If the `pk` is the unique identifier, I'd recommend using the [`defaults`](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#get-or-create) argument.

ons_alpha/datasets/models.py
```diff
@@ -0,0 +1,91 @@
...
+    @property
+    def formatted_edition(self):
+        return self.edition.replace("-", " ").title()
+
+    def __str__(self):
+        return self.title
+
+
+class Dataset(models.Model):
+    id = models.CharField(max_length=255, primary_key=True)
```
RealOrangeOne: Nit: I'm a little concerned about using a manually-constructed value as the primary key. Could we fall back to integers and just set some fields to be unique together?

ons_alpha/datasets/blocks.py
```diff
@@ -0,0 +1,49 @@
...
+    )
+
+    class Meta:
+        template = "templates/components/streamfield/datasets_block.html"
+
+    def clean(self, value):
+        cleaned_value = super().clean(value)
+
+        # Check for duplicate datasets selected through the chooser
+        chosen_datasets = [child.value for child in cleaned_value if isinstance(child.block, DatasetChooserBlock)]
```
RealOrangeOne: Suggestion: You should be able to use `blocks_by_name` to avoid this loop yourself:

```suggestion
        chosen_datasets = cleaned_value.blocks_by_name("dataset_lookup")
```

ons_alpha/core/blocks/related.py
MebinAbraham: Done

ons_alpha/datasets/views.py
MebinAbraham: Doesn't look like it since Queryish doesn't implement all APIs of QuerySet.

ons_alpha/datasets/views.py
MebinAbraham: Things like title/description could change. Fields used to uniquely identify shouldn't change. Addressed.

ons_alpha/datasets/models.py
MebinAbraham: Dropped the custom primary key and used unique together.

ons_alpha/datasets/blocks.py
MebinAbraham: `blocks_by_name` returns list of StreamValue, but I need the value field of each of those. A loop was still needed, but still used this approach over `isinstance`.

ons_alpha/datasets/views.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class DatasetChosenView(ChosenViewMixin, ChosenResponseMixin, View):
+    def get_object(self, pk):
+        # get_object is called before get_chosen_response_data
+        # and self.model_class is Dataset, so we get or create the Dataset from ONSDatasets here
+        # create the dataset object from the API response
+        item = ONSDataset.objects.get(pk=pk)
+        dataset, _ = Dataset.objects.get_or_create(
+            namespace=item.id,
```
RealOrangeOne, line 87: Question: `id` to me looks like it should be a unique identifier - is that not correct for `OnsDataset`?

RealOrangeOne (approved):

ons_alpha/datasets/views.py
zerolab: see https://api.beta.ons.gov.uk/v1/datasets

`id` is unique, but datasets have editions and versions, so we so you can end up with several with the same id. See https://api.beta.ons.gov.uk/v1/datasets/wellbeing-local-authority/editions/time-series/versions

PR #33, opened 2024-08-02
Surface bundle permissions

RealOrangeOne:

Bundles can have permissions assigned, but this isn't exposed to groups.

### How to review

image

### DIFF
--- a/ons_alpha/bundles/wagtail_hooks.py
+++ b/ons_alpha/bundles/wagtail_hooks.py
@@ -1,3 +1,4 @@
+from django.contrib.auth.models import Permission
 from django.urls import include, path
 from django.utils.timezone import now
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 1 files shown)


zerolab (approved):
LGTM.

@MebinAbraham for reference https://docs.wagtail.org/en/stable/reference/hooks.html#register-permissions and https://docs.wagtail.org/en/stable/topics/permissions.html#adding-custom-permissions

MebinAbraham (commented):
What am I missing. `Can view` doesn't works as a standalone perm?. If I have a role with "Can View" bundle but not Add/Change/Remove, I still can't "view" a bundle?. It only works if I also give at least one of Add/Change/Delete.

zerolab:
@MebinAbraham that is fixed in 6.2 [release notes](https://docs.wagtail.org/en/stable/releases/6.2.html#other-features)

> Allow access to snippets and other model viewsets to users with “View” permission (Sage Abdullah)

PR #36, opened 2024-08-05
Read-only workflow task

zerolab:
This PR adds a read-only workflow task.

To test locally:

1. create a new read only "Read-only Group approval task" at `/admin/workflows/tasks/index/`
2. update your workflow at `admin/workflows/list/` and add a new step, assigning it to the needed group

Note: used a monkey-patch for the page edit view to prevent edits, but handle comments. This will not be needed once https://github.com/wagtail/wagtail/pull/6025 gets in (ETA Wagtail 6.3 in November)

Screencast: https://www.loom.com/share/f350287027a24a2d921edb961dd2ccc9

Dev walkthrough: https://www.loom.com/share/899ff363842248808ca1af072a5a80f7


ons_alpha/workflows/static/workflows/css/readonly-task.css
```diff
@@ -0,0 +1,13 @@
...
+    pointer-events: initial!important;
+}
+
+.content-locked.ons-readonly .w-panel__content .DraftEditor-editorContainer {
+    user-select: initial;
+    -moz-user-select: initial;
+    -webkit-user-select: initial;
+    pointer-events: painted;
+}
+
```
zerolab, line 13: When a workflow step is locked for a user, all fields get their pointer events disabled, and selection too. Plus they cannot submit the form.

For ouur read-only step, we allow approve/reject as well as adding comments (which are saved when the draft is saved)

ons_alpha/workflows/static/workflows/js/action-readonly-review.js
```diff
@@ -0,0 +1,59 @@
+function initReviewButton() {
```
zerolab, line 1: this makes the "review" button trigger the popup so the user can approve/reject

ons_alpha/workflows/static/workflows/js/action-readonly-review.js
```diff
@@ -0,0 +1,59 @@
...
+}
+
+function preventContentChanges(editorInstance) {
+  const currentContent = editorInstance.getEditorState().getCurrentContent();
+  const originalOnChange = editorInstance.onChange;
+
+  editorInstance.onChange = (editorState) => {
+    if (editorState.getCurrentContent() !== currentContent) {
+      // Prevent content changes by reverting to the original content
+      return originalOnChange(DraftJS.EditorState.undo(editorState));
```
zerolab: this basically undoes any changes to the rich text editor since we no longer prevent pointer/selection events. Helpful for adding comments

MebinAbraham (approved):

PR #40, opened 2024-08-15
Simple translation

zerolab:
Enables and configures i18n.
- uses a path prefix for the language
- adds a language switcher
- adds some string translations (extracted from our apps/templates)

Screencast: https://www.loom.com/share/064b91f042ec4968af5ddbb69f9fe3af?sid=5b8efee8-ecb1-4133-a427-0aba43649d42
Dev walkthrough: https://www.loom.com/share/0a8bf188e24a4b4fa9d40cd045c65df4?sid=93cc4bfc-eb0b-4ff5-a0dc-3bf0cf3e7a84


MebinAbraham (changes requested):
Few things to things to address:

1. The README should be updated to document how to generate the .po file and then compile to get the .mo file.
2. Since the translations create new pages that represent the same thing, the topic page now errors due to our validation of 'Topic section page with this Topic already exists.'. We would need would need to handle this this for translated pages unless we allow translating topic's themselves. How have you normally handled situation like this before, especially 1:1 relationships?.
3. Current previewing of bulletings are erroring on to the following lines https://github.com/ONSdigital/dis-wagtail-alpha/pull/40/files#diff-107bebabb3ec5a591b296fe91fe34ba6906fb78d5dac4639722124b3949c4b96R48-R49
    - variant is returning None, hence falling over when using `get_url` on `NoneType`.

RealOrangeOne:
I wasn't able to find a solution for changing slug breaking page serving. The links are still correct, but the page isn't accessible at the new slug, only the old. I'm not especially familiar with Wagtail's internationalisation internals, to know where the bug in this might lie, so I'd suggest waiting for Dan's return.

.editorconfig
```diff
@@ -12,3 +12,6 @@ trim_trailing_whitespace = false
 
 [*.{yaml,yml,json,md}]
 indent_size = 2
+
+[Makefile]
+indent_style = tag
```
MebinAbraham:
```suggestion
indent_style = tab
```

zerolab:
So the issue in the S&T was with the fact that the Welsh tree root page (homepage) was unpublished. Once you publish it, all works. Definitely a Wagtail peculiarity.

Now going down into the weeds debugging Wagtail core.

Makefile
```diff
@@ -32,3 +32,9 @@ format-html:  ## 🎨️   - Format the HTML code
 
 build-docker:
 	docker build -t ons_alpha .
+
+messages:
+	python -m manage makemessages --all
```
MebinAbraham:
```suggestion
	poetry run python -m manage makemessages --all
```

Makefile
```diff
@@ -32,3 +32,9 @@ format-html:  ## 🎨️   - Format the HTML code
...
+
+compile-messages:
+	python -m manage compilemessages --use-fuzzy
```
MebinAbraham:
```suggestion
	poetry run python -m manage compilemessages --use-fuzzy
```

MebinAbraham (approved):

PR #41, opened 2024-08-15
Use single cache for renditions and default cache

RealOrangeOne:

The intention is to use a Redis cluster, which doesn't play nicely with multiple databases when using `redis-py`. Using a single Redis database won't cause any issues, and will perhaps save some resources by only using a single connection.

### How to review

Basic smoke test.

Deployments will cache-miss for a while, but otherwise there should be no impact.


zerolab (approved):
LGTM.

For context, this is what https://docs.wagtail.org/en/stable/advanced_topics/images/renditions.html#caching-image-renditions talks about

PR #42, opened 2024-08-15
Allow marking an application deployment as read-only

RealOrangeOne:

The user-facing application should be read-only, without write-access to the database, or any access to the Wagtail admin.

This PR adds a `READ_ONLY_ENV` environment variable which, when set, marks the environment as read-only and disables a large amount of functionality:

- The Wagtail admin interface is uninstalled and inaccessible
- The Django admin interface is uninstalled and inaccessible
- Sessions are disabled (neither created nor read)
- Django's `messages` framework is disabled (it's not used on the public interface, and removing it boosts performance a little)
- Authentication backends are disabled (belt-and-braces protection against surprise logins)

The biggest omission from this PR is that admin-based static files from some apps are still available to the read-only application. It's possible to filter these per-file if needed, however this can be quite brittle, and would need constantly reviewing with new package updates. Given that, I've left it for future work.

At this time, a writable DB connection is still needed.

### How to review

Try setting `READ_ONLY_ENV` and confirming public functionality still works, and the admin isn't accessible.

Also try using the read-only DB user.

### DIFF
--- a/docker/postgres-init-dev.sql
+++ b/docker/postgres-init-dev.sql
@@ -1,3 +1,4 @@
 -- Create user for testing read-only connections (DATABASE_URL=postgres://ons_alpha_read_only@db/ons_alpha)
 CREATE ROLE ons_alpha_read_only WITH LOGIN;
 GRANT SELECT ON ALL TABLES IN SCHEMA public to ons_alpha_read_only;
+GRANT INSERT, UPDATE ON wagtailimages_rendition, wagtailforms_formsubmission TO ons_alpha_read_only;
--- a/docker/postgres-init-dev.sql
+++ b/docker/postgres-init-dev.sql
@@ -1,0 +1,3 @@
+-- Create user for testing read-only connections (DATABASE_URL=postgres://ons_alpha_read_only@db/ons_alpha)
+CREATE ROLE ons_alpha_read_only WITH LOGIN;
+GRANT SELECT ON ALL TABLES IN SCHEMA public to ons_alpha_read_only;
--- a/docker/postgres-init-dev.sql
+++ b/docker/postgres-init-dev.sql
@@ -1,4 +1,0 @@
--- Create user for testing read-only connections (DATABASE_URL=postgres://ons_alpha_read_only@db/ons_alpha)
-CREATE ROLE ons_alpha_read_only WITH LOGIN;
-GRANT SELECT ON ALL TABLES IN SCHEMA public to ons_alpha_read_only;
-GRANT INSERT, UPDATE ON wagtailimages_rendition, wagtailforms_formsubmission TO ons_alpha_read_only;
--- a/ons_alpha/settings/base.py
+++ b/ons_alpha/settings/base.py
@@ -129,7 +129,7 @@ MIDDLEWARE = [
     # According to the official documentation it should be listed underneath
     # SecurityMiddleware.
     # http://whitenoise.evans.io/en/stable/#quickstart-for-django-apps
-    "whitenoise.middleware.WhiteNoiseMiddleware",
+    "ons_alpha.utils.whitenoise.ONSWhiteNoiseMiddleware",
     "django.middleware.locale.LocaleMiddleware",
     "django.middleware.common.CommonMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
--- a/ons_alpha/settings/base.py
+++ b/ons_alpha/settings/base.py
@@ -4,6 +4,7 @@ Django settings for ons_alpha project.
 
 import os
 import sys
+import warnings
 
 from pathlib import Path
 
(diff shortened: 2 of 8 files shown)


docker/postgres-init-dev.sql
```diff
@@ -0,0 +1,3 @@
+-- Create user for testing read-only connections (DATABASE_URL=postgres://ons_alpha_read_only@db/ons_alpha)
+CREATE ROLE ons_alpha_read_only WITH LOGIN;
+GRANT SELECT ON ALL TABLES IN SCHEMA public to ons_alpha_read_only;
```
zerolab: q: what about renditions? when you try to access a rendition for the first time, it will create a `Rendition` object..

zerolab (commented):
Tested locally and works well. 
The renditions question remains

docker/postgres-init-dev.sql
RealOrangeOne: Oh, that's a problem. We might need to selectively grant rendition write access, although renditions would have to be stored in the same bucket regardless of how we're configuring S3, to make sure the file always exists if the `Rendition` does.

ons_alpha/urls.py
```diff
@@ -15,21 +13,27 @@
...
-private_urlpatterns = []
+# Private URLs are not meant to be cached.
+private_urlpatterns = [
+    path("documents/", include(wagtaildocs_urls)),
+]
+
+if apps.is_installed("wagtail.admin"):
+    from wagtail.admin import urls as wagtailadmin_urls
+
+    private_urlpatterns.append(path("admin/", include(wagtailadmin_urls)))
```
RealOrangeOne, line 27: Thought: Should the writable instance be _only_ the admin, and not allow serving the regular site? Perhaps just redirecting to the admin?

MebinAbraham: Not serving the regular would make local development a pain, as we would have to run both and update the routing to use the read-only app. I don't see an added risk of allowing the regular site via admin instance. I want to reduce the no. of moving parts/variants in delivering this solution. We can block the regular site from admin instance if there is a clear need.

docker/postgres-init-dev.sql
MebinAbraham: Can we not pre-generate them instead of on-demand?

RealOrangeOne: Unfortunately not.

As this needs more thinking, I've extracted it out into a separate ticket.

MebinAbraham (commented):
Currently depending on the deployed mode, different migrations are being applied.

When read only false:
```
Migrations for 'wagtailcore':
  /site-packages/wagtail/migrations/0098_delete_admin_delete_editingsession.py
    - Delete model Admin
    - Delete model EditingSession
```
When Read only true:
```
Migrations for 'wagtailcore':
  /site-packages/wagtail/migrations/0099_admin_editingsession.py
    - Create model Admin
    - Create model EditingSession
```

I can see this leading to race conditions given the underlying DB is the same, plus if it was truly read-only, DB migrations can't be applied?. Currently, they conflict, but we need to find a way in which read-only or not doesn't change migrations.

RealOrangeOne:
The read-only app shouldn't ever create migrations, but It's strange there's a difference. I'll see if I can figure out why and rationalise.

ons_alpha/settings/base.py
```diff
@@ -119,18 +127,37 @@
...
+    "wagtail.contrib.settings.context_processors.settings",
+    # This is a custom context processor that lets us add custom
+    # global variables to all the templates.
+    "ons_alpha.core.context_processors.global_vars",
+]
+
+if not READ_ONLY_ENV:
+    context_processors.extend(
+        ["django.contrib.messages.context_processors.messages", "django.contrib.auth.context_processors.auth"]
+    )
```
MebinAbraham: Given it is always installed, keep this enabled at all times for audit purposes?

ons_alpha/urls.py
```diff
@@ -15,21 +13,29 @@
...
+]
+
+# `wagtail.admin` must always be installed,
+# so check `READ_ONLY_ENV` directly.
+if not settings.READ_ONLY_ENV:
+    from wagtail.admin import urls as wagtailadmin_urls
+
+    private_urlpatterns.append(path("admin/", include(wagtailadmin_urls)))
+
+if apps.is_installed("django.contrib.admin"):
```
MebinAbraham, line 29: Minor, worth treating this the same as above. `not settings.READ_ONLY_ENV:` Regardless of if it is installed, if it is read-only, then don't allow. If we ever need to enable it for a read-only env for debug purposes, it can be extended.

MebinAbraham (approved):

ons_alpha/settings/base.py
RealOrangeOne: Both of these won't be used in the frontend, and just add extra overhead to template rendering. Removing them should bring a small performance improvement

ons_alpha/urls.py
RealOrangeOne: I'd much rather be using `is_installed` - Wagtail gets unhappy if `wagtail.admin` isn't installed, hence needing to use `READ_ONLY_ENV`. Having as few special-case uses of `READ_ONLY_ENV` as possible makes things easier to follow, especially when deriving functionality based on when an app is installed.

PR #43, opened 2024-08-22
Enforce collection restrictions on images

RealOrangeOne:

Unlike Documents, Wagtail doesn't enforce collection view restrictions on images. 

This PR forces images to route through a given view, which does these checks.

The checks work for the `{% image %}`, the `{% image_url %}` tag, as well as any manual retrievals of the rendition's URL through its `.url`.

See also https://docs.wagtail.org/en/stable/advanced_topics/images/image_serve_view.html

### How to review

Creating a collection with the various restrictions, and confirm they're enforced.

Note that unlike documents and pages, this PR doesn't add the ability to fix the permissions issue (eg redirect to a password or login screen).

### DIFF
--- a/ons_alpha/images/models.py
+++ b/ons_alpha/images/models.py
@@ -1,5 +1,22 @@
+from typing import NamedTuple
+
+from django.core.files.base import ContentFile
 from django.db import models
 from wagtail.images.models import AbstractImage, AbstractRendition, Image
+
+
+class StubImage(NamedTuple):
+    """
+    A stub image class for use when generating rendition image URLs.
+    """
+
+    id: int
+
+    @property
+    def file(self):
+        # HACK: Provide a blank filename to satisfy `reverse`,
+        # but the view never uses it.
+        return ContentFile("", name="")
 
 
 class CustomImage(AbstractImage):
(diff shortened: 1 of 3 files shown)


ons_alpha/images/models.py
```diff
@@ -1,7 +1,24 @@
...
+    A stub image class for use when generating rendition image URLs.
+    """
+
+    id: int
+
+    @property
+    def file(self):
+        # HACK: Provide a blank filename to satisfy `reverse`,
+        # but the view never uses it.
+        return ContentFile("", name="")
```
MebinAbraham, line 19: What if we did want to generate actual rendition URLs?. I think this needs a bit of clarity on why we are doing this.
What the scenarios that currently fail without this, and what may the actual solutions look like?.

MebinAbraham (commented):
> Note that unlike documents and pages, this PR doesn't add the ability to fix the permissions issue (eg redirect to a password or login screen).

Can you elobrate?

MebinAbraham (commented):
I think this needs a further convo, although this allows the images to be restricted based on certain criterions, the fundamental need of any uploaded assets shouldn't be public until publish time still persists. Right now the uploaded docs/images are independent of the page publish, so they are accessible even if the page isn't published. We need to understand what it takes to make them be tied to the status of a page etc.

zerolab:
Closing in favour of #58

PR #44, opened 2024-08-23
Update topic model and integrate with Design System NWP-256

AWitcherONS:
To Update the topic models add the design system and add the child pages indexes

### How to review
To be added


ons_alpha/jinja2/templates/pages/topics/topic_page.html
```diff
@@ -1 +1,47 @@
...
+            <div class="ons-grid__col ons-col-8@m ons-col-12@s ons-u-p-no@xxs@m">
+                <section id="summary">
+                    <h2>{{ ("Summary")}}</h2>
+                    <p>{{ page.summary }}</p>
+                </section>
+            </div>
+        </div>
+    {% endblock %}
+
+    {% block topic_page_nav %}
```
RealOrangeOne: Issue: This block names feels too specific for the base template. Is there a better name for it? `preSubpages` perhaps?

Alternatively, should this nav section be on every page, or just topic pages? Perhaps it could optionally be shown if `subpages_dict` (or another value) is defined?

ons_alpha/jinja2/templates/pages/topics/topic_page.html
```diff
@@ -1 +1,47 @@
 {% extends "templates/pages/index_page.html" %}
+
+{% from "components/panel/_macro.njk" import onsPanel %}
+
+{% block main %}
+    {% block summary %}
```
RealOrangeOne: Nit: I don't think you need to nest blocks like this - they'll be overridden regardless of where they are. This way it doesn't accidentally override everything in the `main` block.

ons_alpha/jinja2/templates/pages/topics/topic_page.html
```diff
@@ -1 +1,47 @@
...
+        <div class="ons-grid ons-js-toc-container ons-u-ml-no ons-u-mt-s">
+            <div class="ons-grid__col ons-col-8@m ons-col-12@s ons-u-p-no@xxs@m">
+                <section id="summary">
+                    <h2>{{ ("Summary")}}</h2>
```
RealOrangeOne: Suggestion: If the intention is to make this title translatable, it needs to call the `_` method, not just wrap the string in brackets:

```suggestion
                    <h2>{{ _("Summary")}}</h2>
```

ons_alpha/jinja2/templates/pages/topics/topic_page.html
```diff
@@ -1 +1,47 @@
...
+    {% endblock %}
+
+    {% block index %}
+        <div class="ons-grid ons-js-toc-container ons-u-ml-no ons-u-mt-s">
+            <div class="ons-grid__col ons-col-8@m ons-col-12@s ons-u-p-no@xxs@m">
+                {% if subpages %}
+                    <h2>{{"Index"}}</h2>
+                    {% for subpage_key in subpage_dict.keys() %}
+                        <h3 id="{{ subpage_key }}">{{ subpage_key }}</h3>
+                        {% for subpage_value in subpage_dict[subpage_key]  %}
```
RealOrangeOne: Suggestion: You can do `{% for subpage_key, subpage_key in subpage_dict.items() %}` to avoid the need for nested loops.

ons_alpha/jinja2/templates/pages/topics/topic_page.html
```diff
@@ -1 +1,47 @@
...
+                            <a href="{{ subpage_value.url }}">
```
RealOrangeOne: Issue: For getting the URL of a page, it's best to use the [`{% pageurl %}`](https://docs.wagtail.org/en/stable/topics/writing_templates.html#pageurl) tag, as it's significantly faster, especially in loops.

ons_alpha/jinja2/templates/pages/topics/topic_section_page.html
```diff
@@ -1 +1,40 @@
 {% extends "templates/pages/index_page.html" %}
+
+{% from "components/panel/_macro.njk" import onsPanel %}
+
+{% block main %}
+    {% block summary %}
+        <div class="ons-grid ons-js-toc-container ons-u-ml-no ons-u-mt-s">
+            <div class="ons-grid__col ons-col-8@m ons-col-12@s ons-u-p-no@xxs@m">
+                <section>
+                    <h2>{{"Summary"}}</h2>
```
RealOrangeOne: Issue: Should this be translated?

ons_alpha/jinja2/templates/pages/topics/topic_section_page.html
```diff
@@ -1 +1,40 @@
...
+
+    {% endblock %}
+{% endblock %}
+
+
+
+
+
+
+
```
RealOrangeOne: Issue: trailing lines
```suggestion
```

ons_alpha/jinja2/templates/pages/topics/topic_section_page.html
```diff
@@ -1 +1,40 @@
...
+                    <h2>{{"Summary"}}</h2>
+                    <p>{{ page.summary }}</p>
+                </section>
+            </div>
+        </div>
+    {% endblock %}
+    {% block index %}
+        <div class="ons-grid__col ons-grid__col--sticky@m ons-col-4@m ons-u-p-no">
+            {% if subpages %}
+                <h2>{{"Index"}}</h2>
```
RealOrangeOne: Issue: Translation?

ons_alpha/topics/migrations/0008_topicpage_topic_page_nav.py
RealOrangeOne, line 1: Issue: Please prune some of these migrations so there's only 1 migration file. This makes deploying and reviewing significantly simpler.

ons_alpha/topics/models.py
```diff
@@ -41,19 +44,81 @@ class TopicPage(BaseTopicPage):
     parent_page_types = ["topics.TopicSectionPage"]
     subpage_types = ["articles.ArticleSeriesPage", "bulletins.BulletinSeriesPage", "methodologies.MethodologyPage"]
     page_description = "A specific topic page. e.g. Public sector finance or Inflation and price indices"
+    summary = models.CharField(blank=True, max_length=255)
+    topic_page_nav = models.CharField(blank=True, max_length=255)
```
RealOrangeOne: Issue: It's hard to tell from just the name of this field what it controls. Perhaps some help text? Or a more descriptive name.

ons_alpha/topics/models.py
```diff
@@ -41,19 +44,81 @@ class TopicPage(BaseTopicPage):
...
     def clean(self):
         super().clean()
         if TopicSectionPage.objects.filter(topic=self.topic).exists():
             raise ValidationError({"topic": "Topic Section Page with this topic already exists."})
 
+    def topic_list(self):
+        menu_items = self.get_children().live().public().specific()
+        display_names = []
+        for m in menu_items:
+            display_names.append(m.specific.page_type_display_name )
```
RealOrangeOne: Issue: We shouldn't be using the internal page model names for public use.

ons_alpha/topics/models.py
```diff
@@ -41,19 +44,81 @@ class TopicPage(BaseTopicPage):
...
+            ObjectList(BasePage.promote_panels, heading="Promote"),
+        ]
+    )
 
     def clean(self):
         super().clean()
         if TopicSectionPage.objects.filter(topic=self.topic).exists():
             raise ValidationError({"topic": "Topic Section Page with this topic already exists."})
 
+    def topic_list(self):
```
RealOrangeOne: Issue: What is this method returning? It seems to be the type of pages (topics or sections) contained in this topic? Could this be named better?

ons_alpha/topics/models.py
```diff
@@ -41,19 +44,81 @@ class TopicPage(BaseTopicPage):
...
+        menu_items = self.topic_list()
+        display_names = {}
+        for m in menu_items:
+            display_names[m] = []
+            subpage_list = []
+            for subpage in self.get_children().live().public().specific():
+                if subpage.specific.page_type_display_name == m:
+                    subpage_list.append(subpage)
+            display_names[m] = subpage_list
+        return display_names
```
RealOrangeOne: Question: What is this method trying to do, and why? Is there a reason the pages need to be grouped by type?

ons_alpha/topics/models.py
```diff
@@ -41,19 +44,81 @@ class TopicPage(BaseTopicPage):
...
     page_description = "A specific topic page. e.g. Public sector finance or Inflation and price indices"
+    summary = models.CharField(blank=True, max_length=255)
+    topic_page_nav = models.CharField(blank=True, max_length=255)
+
+    content_panels = BasePage.content_panels + [
+        FieldPanel("summary"),
+        FieldPanel("topic_page_nav"),
+    ]
+
+    edit_handler = TabbedInterface(
```
RealOrangeOne, line 62: Suggestion: This is duplicated from the parent, clearly because it needs access to the `.content_panels`. Perhaps this should be moved to a property on the parent class so it can retrieve the right values.

```python
from django.utils.functional import classproperty

@classproperty
def edit_handler(cls):
    return TabbedInterface(
        [
            ObjectList(cls.content_panels, heading="Content"),
            ObjectList(
                [FieldPanel("topic")],
                help_text="Select the topics that this page relates to.",
                heading="Taxonomy",
            ),
            ObjectList(cls.promote_panels, heading="Promote"),
        ]
    )
```

ons_alpha/jinja2/templates/pages/topics/topic_page.html
AWitcherONS: my nested loop logic
one header for each page type present 
and a list of all the child pages that are of the above page type
(this would then have associated pages (tagged) added to the list by page type)
I'm sorry I can't see any way of removing the second loop

RealOrangeOne: Ah damn, you're right - I'd misead!

However, it might be cleaner to use `.items` regardless:

```
{% for subpage_key, subpage_values in subpage_dict.items() %}
    {% for subpage_value in subpage_values %}
```

ons_alpha/jinja2/templates/pages/topics/topic_page.html
```diff
@@ -1 +1,45 @@
...
+
+    {% block index %}
+        <div class="ons-grid ons-js-toc-container ons-u-ml-no ons-u-mt-s">
+            <div class="ons-grid__col ons-col-8@m ons-col-12@s ons-u-p-no@xxs@m">
+                {% if subpages %}
+                    <h2>{{"Index"}}</h2>
+                    {% for subpage_key in subpage_dict.keys() %}
+                        <h3 id="{{ subpage_key }}">{{ subpage_key }}</h3>
+                        {% for subpage_value in subpage_dict[subpage_key]  %}
+                            <a href="{{ subpage_value.pageurl }}">
```
RealOrangeOne: Not quite:

```suggestion
                            <a href="{% pageurl subpage_value %}">
```
https://docs.wagtail.org/en/stable/topics/writing_templates.html#pageurl

There are also other places where `.url` is used directly, which should be updated.

ons_alpha/topics/models.py
```diff
@@ -41,19 +44,91 @@ class TopicPage(BaseTopicPage):
...
+    def subpage_list(self):
+        child_page_types = self.topic_list_child_page_types()
+        display_names = {}
+        for child_page_type in child_page_types:
+            display_names[child_page_type] = []
+            subpage_list = []
+
+            for subpage in self.get_children().live().public().specific():
+                if subpage.specific.page_type_display_name == child_page_type:
+                    if len(subpage.get_children().live().public().specific()) > 0:
```
RealOrangeOne: Suggestion: There's a more efficient way to determine this. Calling `len` will load all results into a list, and call `len` on it. `.exists` does a very short query in the DB to determine if any rows match.

```suggestion
                    if subpage.get_children().live().public().specific().exists():
```

RealOrangeOne (commented):
Adding a few more comments based on changes, however there are still some unresolved comments which need addressing.

ons_alpha/topics/models.py
RealOrangeOne: I'm not sure calling the result a `subpage_dict` is the best solution either. Is it just meant to be the child topics and topic sections? Dedicated methods to get each is probably better, rather than making it too generic.

ons_alpha/topics/models.py
```diff
@@ -41,17 +41,87 @@ class TopicPage(BaseTopicPage):
     parent_page_types = ["topics.TopicSectionPage"]
     subpage_types = ["articles.ArticleSeriesPage", "bulletins.BulletinSeriesPage", "methodologies.MethodologyPage"]
     page_description = "A specific topic page. e.g. Public sector finance or Inflation and price indices"
+    summary = models.CharField(blank=True, max_length=255)
+    topic_page_nav = models.CharField(blank=True, max_length=255)
```
MebinAbraham: What is topic page nav used for?

ons_alpha/jinja2/templates/pages/topics/topic_page.html
```diff
@@ -1 +1,45 @@
...
+                </section>
+            </div>
+        </div>
+    {% endblock %}
+
+    {% block index %}
+        <div class="ons-grid ons-js-toc-container ons-u-ml-no ons-u-mt-s">
+            <div class="ons-grid__col ons-col-8@m ons-col-12@s ons-u-p-no@xxs@m">
+                {% if subpages %}
+                    <h2>{{"Index"}}</h2>
```
MebinAbraham: Can drop this. Don't need the index title as we have a title for each section.

ons_alpha/jinja2/templates/pages/topics/topic_page.html
```diff
@@ -1 +1,45 @@
...
+        </div>
+    </div>
+
+    {% block preSubpages %}
+        <div class="ons-grid ons-js-toc-container ons-u-ml-no ons-u-mt-s">
+            <div class="ons-grid__col ons-col-8@m ons-col-12@s ons-u-p-no@xxs@m">
+                <section id="topic_page_nav">
+                    {% for key in subpage_dict.keys() %}
+                        <a href="#{{ key }}">{{ key }}</a>
+                    {% endfor %}
```
MebinAbraham: Minor, would be good to show the types that aren't our internal page names like "Article series page" vs "Articles" or "Articles related to {topic}" etc.

ons_alpha/topics/models.py
AWitcherONS: Hard coded with default to humanreadable page type 
https://docs.wagtail.org/en/stable/reference/pages/model_reference.html#wagtail.models.Page.page_type_display_name

ons_alpha/topics/models.py
AWitcherONS: Removed

ons_alpha/jinja2/templates/pages/topics/topic_page.html
AWitcherONS: **Hard coded** with default to humanreadable page type
https://docs.wagtail.org/en/stable/reference/pages/model_reference.html#wagtail.models.Page.page_type_display_name

zerolab:
Marked this as a draft as we merge the alternative (#50)
Leaving open so we compare notes

PR #45, opened 2024-08-29
AWS Deployment

RealOrangeOne:

This PR adds the ability to for the app to be deployed to EKS, based on the [infrastructure](https://github.com/ONSdigital/dis-infra-wagtail-poc-stack/) and [helm charts](https://github.com/ONSdigital/dis-wagtail-alpha-helm).

There are a few notable changes to the application:

- Allow authenticating to the DB using IAM credentials
- Create a dedicated URL for liveness / readiness checks, which hit the DB and cache to confirm they're working correctly.
- Allow connecting to Elasticache using specific variables (IAM auth is possible, but didn't work after lots of poking, so relegated to a future PR).
- Allow using a read replica to separate load for read queries

### How to review

This change should be entirely additive - it shouldn't break existing development, or the Heroku environment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,3 @@
-ECR_AWS_ACCOUNT_ID := $(shell aws sts get-caller-identity --query "Account" --output text)
 DESIGN_SYSTEM_VERSION=`cat .design-system-version`
 
 .DEFAULT_GOAL := help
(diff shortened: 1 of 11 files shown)


helm/values/env-bleed.yaml
```diff
@@ -0,0 +1,48 @@
...
+      value: "wagtail-poc-redis-serverless-rjbupb.serverless.euw2.cache.amazonaws.com"
+    - name: ELASTICACHE_PORT
+      value: "6379"
+    - name: AWS_STORAGE_BUCKET_NAME
+      value: "wagtail-ons-poc-media"
+
+  secrets:
+    - secret: django-secret-key
+      data:
+        secret-key: dnhrJm5ndmdoMzF1d21eN2opK3k5NUB1X25yITB1ZWhwMnNkPTAkeSFAZnA5Mm9saGxeYjVuLXRu
```
RealOrangeOne, line 37: Note: Currently, this value is exposed in the repository. It's currently not used for anything, as the entire environment is protected by IP restriction.

Proper secret management is in progress.

Makefile
```diff
@@ -1,3 +1,6 @@
+TARGET_PROFILE_NAME := dp-bleed-dev
+ECR_PROFILE_NAME := dp-bleed-dev
```
MebinAbraham, line 2: These should be removed. We don't want any reference to our actual infra.

islas104:
Hi @MebinAbraham  @RealOrangeOne  @zerolab ,

I’ve recently changed my phone number and am currently unable to log into Teams or Slack. I’ve tried reaching out to Mike, but haven’t received a response via his mobile, so I’m posting here as a last attempt to seek assistance. My new number is 07429 122400. Could someone please help reset my number so I can regain access?

Thank you for your help.

PR #47, opened 2024-09-05
Add a headline field for bulletins/articles to act as news headlines

zerolab:
This was one of the things that the content transformation team was considering.
Made a quick change to allow it for next week.

cc @MebinAbraham @janderson2


ons_alpha/articles/models.py
```diff
@@ -45,7 +46,15 @@ class ArticlePage(BundledPageMixin, RoutablePageMixin, BasePage):
     updates = StreamField(CorrectionsNoticesStoryBlock(), blank=True, use_json_field=True)
 
     content_panels = (
-        BasePage.content_panels
+        [
+            FieldPanel("title"),
+            FieldPanel(
+                "headline",
+                help_text="Use this as a news headline. When set, replaces the title on the page. "
+                "Note that the page <code>slug</code> is driven by the title field.",
```
RealOrangeOne, line 55: Nit: They can be changed independently. Is this something which needs flagging here explicitly?

RealOrangeOne (approved):

ons_alpha/articles/models.py
zerolab: I am not sure. Let's get feedback and see :)

PR #49, opened 2024-09-06
Bundle level workflow

zerolab:
The commits should be fairly atomic, but as an overview:

- adds an admin dashboard panel to make bundles more visible
  Screenshot 2024-09-06 at 16 29 51
- adds validation to prevent the bundle creator from marking it as ready
- adds a placeholder workflow task that we expand the validation for the approval
- adds a "save and approve" button
- modifies the page chooser to provide more information (like the page type and the workflow stage the page is in. NWP-389)
- adds a "publish" bundle when the approved bundle has no scheduled date
- adds a page summary table in the inspect view (NWP-389)

To test locally:
- update your worklow (settings > workflows) to have the last task as "ready to be published" task

Screencast: https://www.loom.com/share/bccf51fcf78f4d6182553c5b2f54817b


RealOrangeOne (approved):

ons_alpha/bundles/models.py
```diff
@@ -6,33 +6,26 @@
...
+from .panels import BundleNotePanel, PageChooserWithStatusPanel
 
 
 class BundlePage(Orderable):
     parent = ParentalKey("Bundle", related_name="bundled_pages", on_delete=models.CASCADE)
     page = models.ForeignKey("wagtailcore.Page", blank=True, null=True, on_delete=models.SET_NULL)
 
     panels = [
-        PageChooserPanel("page", ["articles.ArticlePage", "bulletins.BulletinPage", "methodologies.MethodologyPage"]),
+        PageChooserWithStatusPanel(
```
MebinAbraham, line 26: Non-blocker, can you filter our live pages out of the box?

MebinAbraham (approved):
Works well for the PoC but as discussed, the process of bundles/workflows needs to be polished for Beta as there are quite a few odd cases at the moment but not important for the alpha. We'll see what the feedback is like and adapt in Beta.

ons_alpha/bundles/models.py
zerolab: This could be done via `construct_page_chooser_queryset` or better, by registering a custom chooser.

Thinking of the logic for this - we will want to filter out live pages that do not have any pending drafts

zerolab: will look at tackling separately

PR #50, opened 2024-09-09
Topic pages: show children

zerolab:
This PR is an alternative to #44 (NWP-256)

- shows the latest bulletins/articles in series for a topic page
- shows last 5 methodology pages, ordered by last revision date
- uses https://service-manual.ons.gov.uk/design-system/components/timeline for each section, but can be updated to use https://service-manual.ons.gov.uk/design-system/components/document-list if that is deemed a better fit

It takes simpler route (IMHO) in that we know what types of pages we have under topic pages

Screencast: https://www.loom.com/share/6ec07dbb57164e2789200ffd793c7399
Dev screencast: https://www.loom.com/share/ffb5516202bd4c08b08b299d38b0d169


MebinAbraham (approved):
Unrelated to this but noticed that preview is broken due to bug in `get_translation_urls` when the page is in draft.

On a related note, this makes it clear we need analysis on how the topic page should be curated, i.e what bulletins/articles to show. Last 5 vs latest per series as you have got it vs a curated choice of series for which the latest is pulled.

zerolab:
@MebinAbraham the preview fix is in #52

I agree. The topic page also hinges on the design. I will happily extend this to allow for curation if you think it will help the simulation exercise

MebinAbraham:
> @MebinAbraham the preview fix is in #52
> 
> I agree. The topic page also hinges on the design. I will happily extend this to allow for curation if you think it will help the simulation exercise

We may as well wait until we see the new design and change for beta.

zerolab:
Added a note about the logic in 07b84c1b2baa2634882f0b5c42e8314284eb9140 (in #51)

PR #53, opened 2024-09-10
Validate topic name uniqueness per parent

zerolab:
hat tip to @AWitcherONS for spotting this

### DIFF
--- a/ons_alpha/taxonomy/models.py
+++ b/ons_alpha/taxonomy/models.py
@@ -77,6 +77,12 @@ class TopicForm(WagtailAdminModelForm):
         if not parent and instance.get_parent() != parent:
             raise ValidationError("Cannot make a child topic a top-level topic.")
 
+        # depth starts with one. We cannot rely on instance.get_parent()
+        # until after the topic is created, and parent can be None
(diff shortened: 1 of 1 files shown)


MebinAbraham (approved):

ons_alpha/taxonomy/models.py
```diff
@@ -77,6 +77,12 @@ def clean(self):
         if not parent and instance.get_parent() != parent:
             raise ValidationError("Cannot make a child topic a top-level topic.")
 
+        # depth starts with one. We cannot rely on instance.get_parent()
+        # until after the topic is created, and parent can be None
+        depth = parent.depth + 1 if parent else 1
+        if Topic.objects.filter(name=cleaned_data["name"], depth=depth).exists():
```
RealOrangeOne, line 83: Would `if self.instance.get_siblings().exclude(id=self.instance.id).filter(name=cleaned_data["name"]).exists()` be better? This avoids the need to calculate the depth.

zerolab: no. because the validation happens before we have an actual saved instance, so things like `.get_parent()`/`.get_siblings()` don't return anything, nor do we have an id

PR #56, opened 2024-09-12
Allow web concurrency to be modified at runtime

RealOrangeOne:

Allow tweaking the number of web workers per container through environment variables.

Config overrides environment variables, for some reason.

### How to review

There is no functional change in this PR, however running gunicorn with a different `WEB_CONCURRENCY` will run that number of processes, rather than just 2.


MebinAbraham (approved):
From the docs, [gunciorn config order](https://docs.gunicorn.org/en/latest/configure.html) (in order of least to most authoritative):

```
Environment Variables
Framework Settings
Configuration File
GUNICORN_CMD_ARGS
Command Line
```

PR #57, opened 2024-09-13
Add front-end tooling so we can use our own JS and CSS in the repo

helenb:
- This PR adds the ability to write custom JS and CSS separately from the design system
- Uses static_src as the source directory for static files, and static_compiled as the target directory for compiled and minimised CSS, JS and images
- Adds `static_src` to the `STATICFILES_DIRS` array (the current `jinja2/assets` folder also still works)
- Adds linting using torchbox set-up
- Adds all the usual kit set-up with webpack and package.json
- Updates the CI to ensure front-end files are compiled and linted
- Updates pre-commit to lint front-end files
- Adds a temporary rule to main.scss so it isn't empty
- Adds node and building of front-end assets to docker

### How to review
- In a local terminal, run `nvm use`, then `npm ci`, then `npm run build`
- Make some changes to `main.js` and `main.scss`.
- Re-run `npm run build` and test the changes on any page on the site
- Run `npm run lint` to test the linting
- Run `npm run start` to test watch mode.
- Run `npm run format` to test prettier
- Try modifying `main.scss` to add `color:red` to the `body` rule and then try committing the change - it should fail.


helenb:
@zerolab I have not currently added the step to install node in the docker set-up, but I can do this if this would be useful.

helenb:
We will also want to add the front-end linting to CI

helenb:
> We will also want to add the front-end linting to CI

This is now done - also added to pre-commit.

helenb:
> @zerolab I have not currently added the step to install node in the docker set-up, but I can do this if this would be useful.

Node and npm run build are now in the docker set-up, as discussed at stand-up.

.eslintignore
```diff
@@ -0,0 +1,7 @@
+node_modules
+coverage
+static_compiled
+venv
+jinja2
+core
+workflows
```
RealOrangeOne: Nit: Can these paths be expanded (eg `/ons_alpha/core`), just in case the directory names are reused elsewhere

.github/workflows/ci.yml
```diff
@@ -30,9 +30,43 @@ jobs:
...
-
+  lint-front-end:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '20'
+      - name: Install npm dependencies
+        run: npm install
```
RealOrangeOne:
```suggestion
        run: npm ci --no-audit
```

Dockerfile
```diff
@@ -1,3 +1,17 @@
+# (Keep the version in sync with the node install below)
+FROM node:20-slim as frontend
+
+# Make build & post-install scripts behave as if we were in a CI environment (e.g. for logging verbosity purposes).
+ARG CI=true
+
+# Install front-end dependencies.
+COPY package.json package-lock.json tsconfig.json webpack.config.js tailwind.config.js  ./
```
RealOrangeOne, line 8: Question: Do these files need loading as part of `npm ci`? The fewer files used here, the better caching will be.

ons_alpha/static_src/images/cssBackgrounds/large-file.svg
RealOrangeOne, line 1: Nit: Could this directory be `spine` case instead? `css-backgrounds` to match other URL structures.

package.json
```diff
@@ -0,0 +1,80 @@
...
+    "test": "jest",
+    "test:watch": "jest --watch",
+    "test:coverage": "jest --coverage",
+    "//[ Accessiblity tests ]//": "",
+    "pa11y": "pa11y-ci --sitemap http://localhost:8000/sitemap.xml --config ./pa11y.config.js || (echo 'Please install pa11y-ci globally using: npm install -g pa11y-ci' && exit 1)",
+    "//[ Build tasks ]//": "",
+    "build": "webpack --mode development --progress",
+    "build:prod": "webpack --mode production",
+    "start": "webpack --mode development --progress --watch",
+    "start:reload": "webpack-dev-server"
```
RealOrangeOne, line 38: Suggestion: This should be added to the `Procfile`.

webpack.config.js
```diff
@@ -0,0 +1,219 @@
...
+
+        // See https://webpack.js.org/configuration/dev-server/.
+        options.devServer = {
+            // Enable gzip compression for everything served.
+            compress: true,
+            static: false,
+            host: '0.0.0.0',
+            // When set to 'auto' this option always allows localhost, host, and client.webSocketURL.hostname
+            allowedHosts: 'auto',
+            port: 3000,
```
RealOrangeOne, line 197: Issue: This port should be exposed in the `docker-compose.yml`.

.eslintignore
helenb: sure - I'll leave ones like `node_modules` which should never be linted even if there was more than one directory.

Dockerfile
helenb: I think so - I copied this over from kit.

ons_alpha/static_src/images/cssBackgrounds/large-file.svg
helenb: I think it might be named like that for webpack. I'll check though.

helenb: I renamed it and it all worked fine.

webpack.config.js
helenb: I've just spotted https://git.torchbox.com/internal/wagtail-kit/-/blob/main/layers/base/docker/docker-compose-frontend.yml?ref_type=heads in kit - do we need that file in this repo too? Is that instead of adding the port to docker-compose.yml? https://git.torchbox.com/internal/wagtail-kit/-/blob/main/layers/base/docker-compose.yml?ref_type=heads#L23

RealOrangeOne: Yes - that lets the frontend run properly in docker.

helenb: @RealOrangeOne  I've just added the port to docker-compose.yml for now - is that sufficient for this project? I never run the fe tooling in docker myself so would be good if you could test if you have a chance.

helenb: Hmm, http://localhost:3000/ doesn't work at all so I'm guessing this isn't correct. Any hints to get this working?

docker/bashrc.sh
```diff
@@ -6,5 +6,12 @@ if [ "$BUILD_ENV" = "dev" ]; then
...
-    alias honcho="honcho -f docker/Procfile"
 fi
+
+# nvm
+if [ -a "$HOME/.nvm/nvm.sh" ]; then
+  export NVM_DIR="$HOME/.nvm"
+  source "$NVM_DIR/nvm.sh"
+fi
+
+alias honcho="honcho -f docker/Procfile"
```
RealOrangeOne: Nit: I think it makes sense to keep these files inside the above `if` - they're only relevant for dev.

webpack.config.js
RealOrangeOne: @helenb did you get this working?

helenb: Yes I did.

docker/bashrc.sh
RealOrangeOne: I'd actually meant this and the `nvm` sourcing - since `nvm` will only be installed in dev anyway.

RealOrangeOne (approved):

PR #58, opened 2024-09-13
Private images and documents

ababic:

Groundwork for using a collections hierarchy with fixed "Public" and "Private" roots for determining the privacy of individual objects, and to determine what ACLs should be set on their related files.

The abstract `PrivateMediaCollectionMember` model is where most of the magic happens. When an individual object is saved, it:
- Tracks any changes to privacy and sets `is_private` and `privacy_last_changed` field values accordingly
- If the object moved from a private to public collection, it sets the `last_public_collection` field value to equal the current `collection` value
- If the object moved from a public to a private collection, it sets the `last_private_collection` field value to equal the current `collection` value

In addition, it also provides methods for:
- Changing an object's `collection` value to the most appropriate 'public' collection
- Changing an object's `collection` value to the most appropriate 'private' collection
- Setting the ACLs for all of an object's related files and recording when they have been updated successfully

And a custom Manager with methods for:
- Bulk-updating the `collection` value of objects to the most appropriate 'public' collection
- Bulk-updating the `collection` value of objects to the most appropriate 'private' collection
- Bulk-updaing ACLS for a list of objects (and recording when they have been updated successfully)

It also includes:
- A data migration to create the 'Public' and 'Private' collections
- A management command to retry failed ACL-setting attempts (run regularity to be confirmed)
- Middleware to prevent editing/deleting of the 'Public' and 'Private' collections in Wagtail
- Some patching of the 'Add to collection' bulk actions that appear in Wagtail on the image and document list views- so, if moving from a public to private (or private to public) collection, the privacy-related fields are updated, and ACLs on related files set accordingly
- Some patching of the Wagtail's 'CollectionForm' to prevent moving of sub-collections between the 'Private' and 'Public' branches. I did have a try at patching the underlying field to prevent selection in the first place, but the view modifies the querysets associated with the field, so it's difficult to get changes to 'stick', and the code was cumbersome as a result


ons_alpha/private_media/models.py
```diff
@@ -0,0 +1,217 @@
...
+
+        If all of the change requests are successful, the object's `acls_last_set`
+        datetime will be updated on completion, and `True` will  be returned.
+
+        If any of the change requests fail, the `acls_last_set` datetime will not
+        be updated, and `False` will be returned.
+        """
+        files = list(self.get_privacy_controlled_files())
+        failures = 0
+        for file in files:
```
RealOrangeOne:
```suggestion
        failures = 0
        for file in self.get_privacy_controlled_files():
```

ons_alpha/private_media/models.py
```diff
@@ -0,0 +1,217 @@
...
+            return False
+
+        self.acls_last_set = timezone.now()
+        return True
+
+    def get_privacy_controlled_files(self) -> Iterable[File]:
+        raise NotImplementedError
+
+    @classmethod
+    def bulk_make_public(cls, objects: list["PrivateMediaMixin"]) -> int:
```
RealOrangeOne: Issue: These feel like methods better assigned to the manager or queryset than the instance itself.

ons_alpha/private_media/models.py
```diff
@@ -0,0 +1,217 @@
...
+    def bulk_set_file_acls(cls, objects: list["PrivateMediaMixin"], acl_name: str) -> int:
+        """
+        For a list of objects of this type, set the ACLs for all related
+        files to either 'public' or 'private'. Returns the number of objects
+        for which all related files were successfully updated - which will
+        also have their 'acls_last_set' datetime updated.
+        """
+        successfully_updated = []
+        for obj in objects:
+            if obj.set_file_acls(acl_name):
```
RealOrangeOne: Question: Does `boto3` have an API to bulk update ACLs? I strongly suspect any performance issues will be from updating the ACLs in S3 rather than updating the DB.

ons_alpha/private_media/models.py
```diff
@@ -0,0 +1,217 @@
...
+        self.was_private_on_init = self.is_private
+
+    @property
+    def is_public(self) -> bool:
+        return not self.is_private
+
+    def determine_privacy(self) -> bool:
+        privacy_changed = False
+        self.is_private = self.collection.path.startswith(get_private_root_collection_path())
+        if self.is_private and not self.was_private_on_init:
```
RealOrangeOne: Nit: `was_private_on_init` seems to only be used here. It might be easier to store `was_private_on_init` in the function before setting `is_private`, since it's only read here to compare to the previous value of `is_private`

ons_alpha/private_media/models.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+class PrivateMediaMixin(models.Model):
+    is_private = models.BooleanField(default=False, editable=False)
+    last_public_collection = models.ForeignKey(
+        Collection, null=True, editable=False, on_delete=models.SET_NULL, related_name="+"
+    )
+    last_private_collection = models.ForeignKey(
+        Collection, null=True, editable=False, on_delete=models.SET_NULL, related_name="+"
+    )
```
RealOrangeOne: Issue: Currently, if a user moves an object into a collection below one of these (say "Private" to "Private / Page 1"), `last_private_collection` isn't updated. It will be if the permission changes, but otherwise it's not. I'm not sure how much of a concern that is, but it's be nice to handle it.

ons_alpha/private_media/models.py
```diff
@@ -0,0 +1,217 @@
...
+                    # TODO: differentiate between general connection failures
+                    # and other issues that couldn't possibly result in success
+                    logger.exception("Failed to set ACL for file: %s.", repr(file))
+                    failures += 1
+                    continue
+
+        if failures:
+            return False
+
+        self.acls_last_set = timezone.now()
```
RealOrangeOne: Nit: if `files` is empty, this is still updated.

ons_alpha/private_media/models.py
ababic: True, but I don't think it's possible to make concrete models inherit the manager from an abstract model class... unless that's changed now?

ons_alpha/private_media/models.py
zerolab: it has batch operations:
https://boto3.amazonaws.com/v1/documentation/api/1.26.103/reference/services/s3control/client/create_job.html
https://aws.amazon.com/blogs/storage/updating-amazon-s3-object-acls-at-scale-with-s3-batch-operations/

ababic: @RealOrangeOne I was hoping for such a thing too, but couldn't find one unfortunately.

ons_alpha/private_media/models.py
ababic: Nice catch! Have updated `determine_privacy()` to set them regardless of a 'change in privacy' now

ons_alpha/private_media/models.py
ababic: I think that's okay. The purpose of `acls_last_set` is to work out whether setting of ACLs still needs to take place. If no files need to be updated at all, there are also no remaining acls to set, so that's fine.

ons_alpha/private_media/models.py
ababic: Okay, moved to a custom Manager class now

ons_alpha/private_media/models.py
ababic: Am going to leave this for now, as triggering batch operations seems way more complicated than it's worth

ons_alpha/private_media/middleware.py
```diff
@@ -0,0 +1,36 @@
...
+                request,
+                "Sorry, the root collections for this project are protected and cannot be edited.",
+            )
+            return redirect("wagtailadmin_collections:index")
+        if getattr(request.resolver_match, "view_name", "") == "wagtailadmin_collections:delete" and int(
+            view_kwargs["pk"]
+        ) in (utils.get_private_root_collection_id(), utils.get_public_root_collection_id()):
+            messages.warning(
+                request,
+                "Sorry, the root collections for this projectare protected and cannot be deleted.",
```
RealOrangeOne:
```suggestion
                "Sorry, the root collections for this project are protected and cannot be deleted.",
```

ons_alpha/private_media/signal_handlers.py
```diff
@@ -0,0 +1,104 @@
...
+
+    for ct_id, id_list in ids_by_ctype.items():
+        model_class = ContentType.objects.get_for_id(ct_id).model_class()
+        id_list = [model_class._meta.pk_field.to_python(id) for id in id_list]
+        private_objects_qs = model_class.objects.filter(id__in=id_list, is_private=True).select_related(
+            "collection", "last_public_collection", "last_private_collection"
+        )
+        if issubclass(model_class, AbstractImage):
+            private_objects_qs = private_objects_qs.prefetch_related("renditions")
+        model_class.objects.bulk_make_public(list(private_objects_qs))
```
RealOrangeOne, line 35: Question: What if an image is used on 2 pages? Currently, the first publish would make the image public - is that intended?

ons_alpha/private_media/signal_handlers.py
```diff
@@ -0,0 +1,104 @@
...
+        )
+        if issubclass(model_class, AbstractImage):
+            public_objects_qs = public_objects_qs.prefetch_related("renditions")
+        model_class.objects.bulk_make_private(list(public_objects_qs))
+
+
+def register_signal_handlers():
+    for page_model in PAGE_MODEL_CLASSES:
+        page_published.connect(publish_page_media_on_publish, sender=page_model, dispatch_uid="publish_media")
+        page_unpublished.connect(unpublish_page_media_on_unpublish, sender=page_model, dispatch_uid="unpublish_media")
```
RealOrangeOne: Suggestion: Can this be done by omitting the `sender` argument instead? Looping over the internal `PAGE_MODEL_CLASSES` feels like it's prone to a race condition, where the signal runs before some classes are loaded.

ons_alpha/private_media/utils.py
```diff
@@ -0,0 +1,131 @@
...
+from django.core.exceptions import ObjectDoesNotExist
+from django.db.models.fields.files import FieldFile
+from storages.backends.s3 import S3Storage
+from wagtail.models import Collection
+
+
+logger = logging.getLogger(__name__)
+
+
+_ROOT_COLLECTION_INFO = Local()
```
RealOrangeOne: Question: Given these collections are intended to be functionally immutable, any reason not to use an `lru_cache` here instead?

ons_alpha/private_media/utils.py
```diff
@@ -0,0 +1,131 @@
...
+        # In environments that don't use S3, return True immediately,
+        # as no futher action is needed, and there's nothing to go wrong!
+        logger.info(f"ACL setting is unnecessary for {file.name}")
+        return True
+
+    obj = file.storage.bucket.Object(file.name)
+    try:
+        obj_acl = obj.Acl()
+    except ClientError as e:
+        logger.debug(f"Failed to retreive ACL for {file.name}: {e!r}")
```
RealOrangeOne, line 101: Issue: Use `logger.exception` to properly capture the exception, rather than trying to stringify it in the message. This also correctly reports to tools like Sentry.

ons_alpha/private_media/middleware.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class PreventRootCollectionChangesMiddleware:
+    """
+    A custom Middleware that prevents editing or deletion of the special 'Private' and
+    'Public' collections used by this project to restrict access to private media.
+    """
+
+    def __init__(self, get_response):
+        self.get_response = get_response
```
RealOrangeOne, line 12: Suggestion: Raise `MiddlewareNotUsed` when in `settings.IS_EXTERNAL_ENV`. This will reduce the extra overhead of a migration.

ons_alpha/private_media/middleware.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def __call__(self, request):
+        return self.get_response(request)
+
+    def process_view(self, request, view_func, view_args, view_kwargs):  # pylint: disable=unused-argument
```
RealOrangeOne, line 17: Question: Could this be done with another monkey-patch, rather than needing a complete middleware for a few small cases.

docker-compose.yml
```diff
@@ -6,6 +6,7 @@ services:
         BUILD_ENV: dev
         POETRY_INSTALL_ARGS: ''
       target: dev
+    env_file: .env
```
RealOrangeOne, line 9: Question: What's the reason for this change?

ons_alpha/private_media/signal_handlers.py
ababic: Thanks for sense-checking @RealOrangeOne! Yes, that is the intended behaviour. By publishing a page, you also publish all media used on that page. It's okay for a draft page to be referencing a mix of private and public media, but problematic for a published page to be referencing private media.

ons_alpha/private_media/signal_handlers.py
ababic: Happy to give that a go!

ons_alpha/private_media/utils.py
ababic: I see what you mean. The main problem I was trying to solve was making these values available cheaply when creating or updating a lot of things in bulk, which this achieves. Despite the context, I can't help but feel reticent about caching things for longer. Let me think on this a bit more.

ons_alpha/private_media/utils.py
ababic: Yes, that is how it was implemented originally... However, I'm concerned about raising exceptions for things that the code simply cannot avoid... Especially when we already have a plan for retrying. I'm not sure it's something that would be useful in the longer term (I think it would just be an annoyance).

ons_alpha/private_media/middleware.py
ababic: @RealOrangeOne I'm not so sure about this one. The data migration creates the collections regardless of environment setup (including in tests), so I'd rather they be protected regardless too. There is no hard requirement for this item being included in `MIDDLEWARE`,  so I feel it might just be clearer to leave it out when not required. Does that make sense to you?

ons_alpha/private_media/middleware.py
ababic: Possibly, but I do really hate the monkey-patching too.... It's way way too easy for things to break in odd, difficult to pin-down ways. Try fudging one of the imports in a patch and see what happens :( Middleware is definitely the lesser of two evils, I think

docker-compose.yml
ababic: I was having trouble getting the Docker container to pick up the `.env` file I created locally (containing the test AWS bucket creds). It only seemed to work after adding this change, then trashing and rebuilding the container.

ons_alpha/private_media/signal_handlers.py
```diff
@@ -0,0 +1,103 @@
...
+        ids_by_ctype[ct_id].append(obj_id)
+
+    for ct_id, id_list in ids_by_ctype.items():
+        model_class = ContentType.objects.get_for_id(ct_id).model_class()
+        id_list = [model_class._meta.pk_field.to_python(id) for id in id_list]
+        public_objects_qs = model_class.objects.filter(id__in=id_list, is_private=False).select_related(
+            "collection", "last_public_collection", "last_private_collection"
+        )
+        if issubclass(model_class, AbstractImage):
+            public_objects_qs = public_objects_qs.prefetch_related("renditions")
```
MebinAbraham, line 56: To clarify this will handle renditions and ensure any generated ones will be made private?

MebinAbraham (commented):
I think there are a few changes we need to make for this to be a suitable solution. IMHO, not a fan of the level of overrides/money patching so we will need to ensure this implementation in Beta is well tested and so that we have confidence in the solution. However, this does get us mostly there.

- We should disable the ability to upload any images to non private collections. We cannot rely on user's selecting the correct collection. Have agreed that tying images/docs to a page's publish status is appropriate.
- The media library should only show "Public" images which can be re-used.
- Optional: Users can only re-use "private" images within the same page. (the scope of this is not clear i.e should this be folder level, release series level etc, will depended on RBAC model we choose).
- Something we may have to do for Beta is, if a page is public but the image/doc's ACL is still set to private for whatever reason (i.e delays, API errors etc), we would want to serve the images/doc via the app.

ons_alpha/private_media/signal_handlers.py
ababic: Yes, `bulk_make_private()` calls each objects `get_privacy_controlled_files()` method, which, for images, includes the source file, and those of any renditions.

ababic:
Thanks for taking a look @MebinAbraham. To address your points:

- We should disable the ability to upload any images to non private collections.

Could we create a separate ticket for this one and make sure we flesh out the requirements fully? My main concern here is that Wagtail allows bulk uploading and individual uploading of media directly into each library, which is convenient. When that is the case, how do we know that the media should be private? What are we expecting in terms of editor numbers and levels of access? Would locking this down not just get in the way of lower-access users uploading media for use on non-sensitive pages?

I wonder if there's scope here to do something at the chooser level, so that when adding media for a specific page (from the page editor interface), we can use the page type and status to select an appropriate collection automatically?

- The media library should only show "Public" images which can be re-used.

Again, this is a tricky one. Wouldn't it be perfectly valid for a user putting together a sensitive page to be able to find media they uploaded to tweak things like descriptive text or a focal point for images?

This is a problem Wagtail already solves by allowing collection permissions to be assigned to specific groups. For example, users in the "Publishing officer" group can see and manage everything in all collections, but a lower-level editor might only be able to see things under "Public". We should try to lean on this as much as possible to avoid more hacks. It's just a case of thinking about what permission groups are needed, and creating them.

- Optional: Users can only re-use "private" images within the same page. (the scope of this is not clear i.e should this be folder level, release series level etc, will depended on RBAC model we choose)

As suggested above, this would be entirely possible with the correctly configured permission groups. There could be a peice of work we could do to automatically create and configure groups for some pages - but, that might not even be needed. I propose we revisit this once we have a clearer idea about the levels of users and their specific needs. 

- Something we may have to do for Beta is, if a page is public but the image/doc's ACL is still set to private for whatever reason (i.e delays, API errors etc), we would want to serve the images/doc via the app.

The idea behind the management command included here is that it would be running in the background to identify any media whose ACLs have not yet been successfully updated to reflect a privacy change, and retry on a regular basis. Once set successfully, the `acls_last_set` value would be updated, and the command would stop retrying for that particular item.

ababic:
@MebinAbraham And regarding the monkey-patches, we are of course aligned on that front. Currently though, we simply don't have a choice. Wagtail is never going to have hooks to allow absolutely anything to be overridden in a declaritive way. Custom concepts like this will always require a patch of sorts.

I do have a pretty reliable method for overriding entire views via Middleware, which is a fair bit more predictable (as can be found in officially supported packages like [wagtail-bynder](https://github.com/torchbox/wagtail-bynder/blob/main/src/wagtail_bynder/middleware.py)). You get a complete alternative view as a testable unit. But, technically, it's still a hack.

MebinAbraham (approved):

zerolab:
💪🏼

PR #59, opened 2024-09-16
Add .python-version to git ignore

ababic:
This file is useful for some editor setups, but shouldn't really be committed to the repo, as it can get in the way of being running fab commands. The project runs in Docker, where the python version is hardcoded elsewhere... pinning it here too means that developers need to install a whole new version of Python just to set up the project, which is overkill.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ node_modules/
 /npm-debug.log
 /.idea/
 .vscode
+.python-version
 coverage
 
 # Distribution / packaging


RealOrangeOne:
As this isn't a file in our default configuration, it might be better to manage this yourself in the `.git/info/exclude` file.

MebinAbraham (commented):
We would want to keep this to ensure devs are using the same Python version and part of our flow when using Pyenv.. We will also need to support setting up the project without docker (backing service can be in docker) for Beta. For example, I run this via PyCharm with redis/postgres in Docker.

Therefore, I don't suggest removing it but excluding it in your local env. Fabric should be part of the main dependencies or replaced with Makefile. One we will be consider for Beta.

PR #60, opened 2024-09-17
Add logging to `publish_bundles` command

RealOrangeOne:

This helps extract more accurate timing information.

`logfmt` is used to help with extracting values.

This PR also ensures that if a single bundle fails, future bundles may still be published (or at least attempted)


zerolab (approved):
LGTM

PR #65, opened 2024-10-02
Fix: Image list crashing with ValueError

ababic:
Resolves: https://torchbox.sentry.io/issues/5939713253/events/616114a170b54ec0a36c00f41330b527/?project=4507498919755776

Moves datetime comparison logic to a separate `acls_are_up_to_date()` method, so that None values can be accounted for.

The problem stems from the fact that new fields were added to the model, so existing images exist with `None` values for some of the automatically-managed datetimes. This was missed in local development, because it was carried out in a newly-built environment without any existing objects.


RealOrangeOne (approved):

zerolab (approved):
Makes sense to me

PR #70, opened 2024-10-03
Hard-coded header and navigation

helenb:

Hard-coded header for the November prototype.
Duplicates the header macro from the design system and replaces `serviceLinks` with `navLinks`
Hard-codes the configuration for the navLinks in `base.html`

### How to review
- Check out the `feature/november-prototype-header` branch
- Open any page in a local build, and check that the header displays as per the designs at desktop and mobile
- Ensure that there is only one link in the header navigation - for now, this has a dummy value. We need to hard-code a link to a bulletin page at some point.


.design-system-version
```diff
@@ -1 +1 @@
-70.0.8
+70.0.17
```
helenb: This update is now in `main`

ons_alpha/jinja2/templates/base.html
```diff
@@ -67,6 +278,32 @@
     {% endif %}
 {% endblock %}
 
+{% block header %}
+    {{
+    onsHeaderNew({
+    "wide": pageConfig.wide,
+    "fullWidth": pageConfig.fullWidth,
```
ababic, line 285: Indentation on this section looks a little unclear

.design-system-version
ababic: @helenb This change has also been merged into the `november-prototype` branch - so, a rebase should make this redundant.

ababic (changes requested):
@helenb Just a couple of minor considerations from my perspective. Otherwise looking great!

ons_alpha/jinja2/templates/base.html
helenb: I think this was the linter's doing unfortunately, but I'll check again.

helenb: Nah I can't get it to let me indent it nicely. sorry.

helenb:
@ababic I couldn't get the linter to behave with the indenting - are you happy to approve anyway?

ababic (approved):
LGTM!

PR #71, opened 2024-10-04
Update the design system version for the project

helenb:
Updates the build to use the latest version of the design system templates

### How to review
Check that the css loaded in the header is https://cdn.ons.gov.uk/sdc/design-system/70.0.17/css/main.css


helenb:
@zerolab maybe we can discuss this MR today if you have time. After running `fab build` this change is not actually updating the CSS for me.

I am wondering if now we have front-end tooling set up, we could consider installing with npm instead of the current system: https://service-manual.ons.gov.uk/design-system/guidance/install-with-npm

RealOrangeOne (approved):

PR #73, opened 2024-10-08
Fix: pylint complaints on november-prototype branch

ababic:
CI isn't always running properly for PRs targeting the `november-prototype` branch, so these were missed before. But, these fixes are needed to unblock other MRs


zerolab:
@ababic please add the Nov prototype branch to https://github.com/ONSdigital/dis-wagtail-alpha/blob/main/.github/workflows/ci.yml#L4-L7 ;)

PR #79, opened 2024-10-09
Disclaimer notice

helenb:

Styles the disclaimer notice that will appear on the November prototype pages. I have some queries around the design which are on the [ticket](https://jira.ons.gov.uk/browse/NWP-453), but this will do for a first pass and I can adjust if needed.

Update: The mobile styles now reflect the design system, rather than the figma design after feedback from Joe. I'll remove the mobile screenshot below as this is now out of date.

### How to review
You can test on a local build with either a topic page or a bulletin page.

### Screenshots
<details>
<summary>Expand to view</summary>

**Desktop:**
Screenshot 2024-10-09 at 12 25 05

</details>


ababic (approved):
Nice work @helenb! Approved.

PR #81, opened 2024-10-09
Prototype: Add 'noindex' tag to all pages (BE)

ababic:
Use the 'SEO_NOINDEX' context variable (added by `core.context_processors.global_vars`) to conditionally include a noindex tag on all pages (whether a Wagtail page or otherwise)

### How to review
- View the home page and inspect the source. You should see `<meta name="robots" content="noindex">` in the `<head>` of the page HTML.
- To remove it, add `SEO_NOINDEX = False` to your `local.py` settings file to override the default and restart
- View the home page and inspect the source again. The `<meta name="robots" content="noindex">` should no longer be present.

### DIFF
--- a/ons_alpha/jinja2/templates/base.html
+++ b/ons_alpha/jinja2/templates/base.html
@@ -3,6 +3,7 @@
 {% from "components/external-link/_macro.njk" import onsExternalLink %}
 
 {% block meta %}
+    {{ super() }}
     {% if SEO_NOINDEX %}
         <meta name="robots" content="noindex">
     {% endif %}
--- a/ons_alpha/jinja2/templates/base.html
+++ b/ons_alpha/jinja2/templates/base.html
@@ -2,6 +2,12 @@
 {% from "component_overrides/header/_macro.njk" import onsHeaderNew %}
 {% from "components/external-link/_macro.njk" import onsExternalLink %}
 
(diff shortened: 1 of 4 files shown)


ons_alpha/jinja2/templates/base.html
```diff
@@ -2,6 +2,12 @@
 {% from "component_overrides/header/_macro.njk" import onsHeaderNew %}
 {% from "components/external-link/_macro.njk" import onsExternalLink %}
 
+{% block meta %}
+    {% if SEO_NOINDEX %}
+        <meta name="robots" content="noindex">
```
helenb, line 8: @ababic I see this is also defined in base_page.html - so this block is just being overridden by what is there and isn't needed as well.

helenb (commented):
Thanks @ababic - all works as expected - we just don't need to add the template change to both base.html and base_page.html

ons_alpha/jinja2/templates/base.html
ababic: @helenb I think the problem is actually that we're not calling `super()` in `base_page.html`. If we do that, this shouldn't be necessary.

helenb: that would work.

helenb (approved):

PR #82, opened 2024-10-10
Chore/disable update notification

zerolab:
also disables the "What's new" banner

https://docs.wagtail.org/en/stable/reference/settings.html#wagtail-enable-update-check
https://docs.wagtail.org/en/stable/reference/settings.html#wagtail-enable-whats-new-banner


ababic (approved):
LGTM. Mebin is demoing today though, so might want to hold back on merging/deploying for now (unless this was a request from him)

PR #84, opened 2024-10-10
Prototype: Restrict headline figure blocks to a single item

ababic:
Currently, it's possible to add multiple instances of ``HeadlineFiguresBlock`` to each page, when we only ever want a single instance to be added. These changes enforce that at the field level, so that editors can only add a single item.

### DIFF
--- a/ons_alpha/bulletins/models.py
+++ b/ons_alpha/bulletins/models.py
@@ -46,7 +46,7 @@ class BulletinPage(BundledPageMixin, RoutablePageMixin, BasePage):
         related_name="+",
     )
     is_accredited = models.BooleanField(default=False)
-    headline_figures = StreamField([("figures", HeadlineFiguresBlock())], blank=True)
+    headline_figures = StreamField([("figures", HeadlineFiguresBlock())], blank=True, max_num=1)
     body = StreamField(CoreStoryBlock(), use_json_field=True)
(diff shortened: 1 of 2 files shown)


ons_alpha/topics/models.py
```diff
@@ -77,7 +77,7 @@ class TopicPage(BaseTopicPage):
     ]
     page_description = "A specific topic page. e.g. Public sector finance or Inflation and price indices"
 
-    headline_figures = StreamField([("figures", HeadlineFiguresBlock())], blank=True)
+    headline_figures = StreamField([("figures", HeadlineFiguresBlock())], blank=True, max_num=1)
```
helenb, line 80: I hadn't realised the headline figures were in the topic page too - do we definitely need them there? They aren't in the design.

ons_alpha/bulletins/models.py
```diff
@@ -46,7 +46,7 @@ class BulletinPage(BundledPageMixin, RoutablePageMixin, BasePage):
         related_name="+",
     )
     is_accredited = models.BooleanField(default=False)
-    headline_figures = StreamField([("figures", HeadlineFiguresBlock())], blank=True)
+    headline_figures = StreamField([("figures", HeadlineFiguresBlock())], blank=True, max_num=1)
```
helenb, line 49: I actually made this change on the styling branch but makes more sense to do it in a separate commit.

helenb (approved):

PR #88, opened 2024-10-14
Design system updates

helenb:
With the major design system release, quite a few styling changes are needed:

- Updates to the header and footer templates to class names and variable names
- Add the logo at the bottom of the footer, as per the updated template (but wrap in an if block so it doesn't display in our version)
- Add a media query mixin to match the sizes used for ons, to make media query use more consistent
- Add a function to work out rem-spacing based on a base 16 size
- Update the `ons-grid--flex-grid` classname to be a top level block (like how `ons-grid-flex` has changed in the design system)
- Update font-sizes and line-heights now we are using base 16
- A few fixes to styles I spotted whilst doing these changes

Note that I also need to update the document block styling but that isn't merged yet, so I'll do it in a separate pull request.

### How to review
All the changes can be tested on a bulletin page. It might help to look at the code changes commit by commit.

This may be helpful when reviewing: https://github.com/ONSdigital/design-system/blob/d3c30b9400c9f0bdfca21f838d82b8890bd1ec05/migration_guides/70.x.x-to-72.0.0-migration-guide.md?plain=1#L196

### DIFF
--- a/ons_alpha/jinja2/component_overrides/footer/_macro.njk
+++ b/ons_alpha/jinja2/component_overrides/footer/_macro.njk
@@ -185,35 +185,37 @@
                                 {% endif %}
                             </div>
                         {% endif %}
-                        {% if not params.footerLogo.logos.logo1 %}
-                            {%
-                                set logo1 = {
-                                    'logoUrl': 'https://cy.ons.gov.uk/' if lang == 'cy' else 'https://www.ons.gov.uk/',
-                                    'logoImage': onsLogo | safe
-                                }
-                            %}
-                            {% set logos = [logo1, params.footerLogo.logos.logo2] %}
-                        {% else %}
-                            {% set logos = [params.footerLogo.logos.logo1, params.footerLogo.logos.logo2] %}
-                        {% endif %}
+                        {% if params.footerLogo %}
+                            {% if not params.footerLogo.logos.logo1 %}
+                                {%
+                                    set logo1 = {
+                                        'logoUrl': 'https://cy.ons.gov.uk/' if lang == 'cy' else 'https://www.ons.gov.uk/',
+                                        'logoImage': onsLogo | safe
+                                    }
+                                %}
+                                {% set logos = [logo1, params.footerLogo.logos.logo2] %}
+                            {% else %}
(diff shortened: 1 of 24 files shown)


ons_alpha/jinja2/templates/base.html
```diff
@@ -404,7 +404,9 @@
     "legal": [
     ],
     "poweredBy": '&nbsp;',
```
ababic: Should we still pass 'poweredBy' here if the component no longer supports it?

ons_alpha/jinja2/component_overrides/footer/_macro.njk
```diff
@@ -152,46 +154,68 @@
...
+                                    set logo1 = {
+                                        'logoUrl': 'https://cy.ons.gov.uk/' if lang == 'cy' else 'https://www.ons.gov.uk/',
+                                        'logoImage': onsLogo | safe
+                                    }
+                                %}
+                                {% set logos = [logo1, params.footerLogo.logos.logo2] %}
+                            {% else %}
+                                {% set logos = [params.footerLogo.logos.logo1, params.footerLogo.logos.logo2] %}
+                            {% endif %}
+
```
ababic, line 197: I don't think we've nailed the logo changes just yet... it feels like something that would benefit from better up-front value prep, and less code here. Let's make a note to come back to it later.

ababic (approved):
I won't claim to understand every single change here, but based on the design system change notes, things appear to be sensible enough. Some bits feel a little bit over-complicated, but nothing that can't be ironed out in future - perfection is unrealistic with such a massive set of changes. If it works for you, let's get it in.

ons_alpha/jinja2/templates/base.html
helenb: Good spot - I have removed.

ons_alpha/jinja2/component_overrides/footer/_macro.njk
helenb: As before, this is nearly all copied from the design system - all I have done is wrap the whole thing in an if block, as there was no other obvious way to exclude the ONS logo from appearing in the footer. Any reworking of these templates needs to happen when we port changes back to the design system, not in the prototype.

PR #93, opened 2024-10-16
Prototype: Add dedicated block for chart embedding

ababic:


helenb:
Not sure if this is ready for review but just had a look at the code and looks good. We'll want to update the basic embed template as it will no longer want the chart styling - but I'm happy to do that separately if you prefer.

helenb (approved):

PR #96, opened 2024-10-18
Prototype: Landing page BE

ababic:
We've decided to hijack the HomePage to use as the 'Landing page' in the prototype designs. This PR adds some content options to allow summary text to be specified for the header area, and also adds a body field with the same content options as the Bulletin page. 

We might add some more options later if needed, but should try to get things as close as possible with existing blocks first.


ons_alpha/home/models.py
```diff
@@ -14,19 +16,19 @@ class HomePage(BasePage):
...
 
     strapline = models.CharField(blank=True, max_length=255)
-    call_to_action = models.ForeignKey(
-        "core.CallToActionSnippet",
-        blank=True,
-        null=True,
-        on_delete=models.SET_NULL,
-        related_name="+",
-    )
+    summary = models.TextField(blank=True)
```
helenb, line 19: Could we also remove the 'strapline' field - I'm not sure what this is for?

helenb:
As discussed the blocks available don't really match up with the design at the moment. For example I can't add an introduction for a document list ('You will be able to explore two pages using this service.'), and I have to add a content label (meta data) for a document.

There's no way to add the button currently, as we know - to discuss with them.

Note the feedback thing is out of scope.

But we can style the header and do our best with the content. I do wonder if we could have a cut-down set of blocks for the content though - having charts etc available is potentially confusing - and if we are demo'ing wagtail it would be good to show how it can be refined to suit the page content. And I also don't want them adding random stuff and expecting it to look tip-top on this page.

ons_alpha/home/models.py
ababic: I think we keep it. The 'Title' of the page will be 'Home', but the strapline will be something more lengthy for the header, e.g. "Welcome to the ONS website preview"

ababic: So, an alternative H1 basically!

helenb (approved):

PR #100, opened 2024-10-18
Allow HRs in richtext by default

ababic:
Required for one-off usage on the prototype landing page


helenb:
If this were a full build I'd want this only on the block it's needed on, but happy to bung it in everywhere for the prototype.

helenb (approved):

PR #104, opened 2024-10-21
Fix-  iframe resize issue

helenb:
There was a intermittent bug with the embedded chart iframes not resizing, which was down to interference from the wagtail chart js.

Note that this is a quick fix - we may want to remove wagtail charts more fully from the november prototype branch.

### How to review
Create a bulletin page with a few charts, and load it several times. Make sure the charts always load at 100% width of the parent.

### DIFF
--- a/ons_alpha/jinja2/templates/pages/bulletin_page.html
+++ b/ons_alpha/jinja2/templates/pages/bulletin_page.html
@@ -151,6 +151,4 @@
     <script async src="{{ static('js/third-party/pym.min.js') }}"></script>
     <script async src="{{ static('js/blocks/embeddable.js') }}"></script>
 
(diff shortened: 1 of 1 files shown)


helenb:
@RealOrangeOne suggests we remove wagtail-charts properly rather than leave it broken, which removing this one line would do. @ababic might you be able to help with that?

ababic:
@helenb 

> @RealOrangeOne suggests we remove wagtail-charts properly rather than leave it broken, which removing this one line would do. @ababic might you be able to help with that?

Yes, I can do that in a follow-up after show and tell. Let's just get this merged if it solves a problem in the immediate term

ababic (approved):

PR #108, opened 2024-10-22
Continue to work on iframe resize issue

helenb:
- We are continuing to see issues with iframes not being resized
- This is another change in addition to the removal of the wagtail charts js (that change is still needed)
- Don't load the script that actually runs the resize asynchronously

### How to review
- Test a bulletin page with charts and see if the iframes resize consistently. Note that after a while you get temporarily blocked from loading the data vis charts.
- This change seems to help but I'm not 100% certain that it will resolve the issue completely yet.


ababic (approved):
LGTM!

PR #111, opened 2024-10-22
Further work on iframe resize issue

helenb:
Another test to see if we can resolve the iframe resize issue - Don't load the pym.js script with async

This pr also makes the following changes:
- Move the embeddable.js code to a class that will get linted


ons_alpha/jinja2/templates/pages/bulletin_page.html
```diff
@@ -148,8 +148,6 @@ <h2 class="heading-2">{{ _("Contact details") }}</h2>
 {% block scripts %}
     {{ super() }}
     <script async src="{{ mathjax() }}"></script>
-    <script async src="{{ static('js/third-party/pym.min.js') }}"></script>
```
ababic, line 153: We can't go ahead with this change I'm afraid, as it's a security risk. Let's just try moving it to the head, so that it's more likely to be loaded by the time the other script runs.

helenb: ah no worries I'll revert back to the local copy. Before trying in the head I'd like to try removing the `async` here. I may also need to add some temporary console.log statements to get a better picture of what is going on on staging.

helenb: This has now been reverted.

ababic (approved):

PR #113, opened 2024-10-23
Document design system overrides more fully

helenb:
- Adds more detailed comments on the design system overrides


ababic (approved):
Great to have this. Thanks @helenb!

PR #116, opened 2024-11-05
GTM embed tweaks

ababic:

As per the latest ticket update on https://jira.ons.gov.uk/browse/NWP-513, the embed isn't working quite as expected. I think this could have something to do with the '-' in the ID being escaped, or possibly the extra URL parameters that were present in the design system code, but not in the version provided originally.


helenb (approved):
Tested and looks good to me. Might be worth double checking about the extra parameters with Arwen.

PR #117, opened 2024-11-06
Prototype: Make phase banner feedback URL configurable (and link logo to home page)

ababic:


helenb (approved):
I've fixed the issue with the feedback link text - rest looks good. I've realised the survey url we've added to the env vars is not a final version, which we don't have yet - so have noted this on the ticket.

PR #123, opened 2024-11-11
Never include the current page in breadcrumb nav

ababic:
[GDS guidance on breadcrumbs](https://design-system.service.gov.uk/components/breadcrumbs/) states:

> The breadcrumbs should start with your ‘home’ page and end with the parent section of the current page.

So, we need to remove our conditional inclusion of the current page, and instead never include it as the final item.

### DIFF
--- a/ons_alpha/articles/models.py
+++ b/ons_alpha/articles/models.py
@@ -19,7 +19,10 @@ from wagtail.models import Page
 from wagtail.search import index
 
 from ons_alpha.bundles.models import BundledPageMixin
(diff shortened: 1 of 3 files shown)


helenb (approved):
Normally I'd expect the template logic to hide the current breadcrumb page - but I guess it is possible in the model code because of the strict page hierarchy. This works fine for the prototype.

helenb:
> Normally I'd expect the template logic to hide the current breadcrumb page - but I guess it is possible in the model code because of the strict page hierarchy. This works fine for the prototype.

ignore this - I misread the template change

PR #126, opened 2024-11-11
Add FeaturedDocumentWithChartBlock for use on TopicPage

ababic:


helenb:
Just had a look at this, and I can see from the design why you chose the featured document block variant - and in fact my argument about the list item applies to the normal featured document block too - you get a list of one in that case (not that I think we should change that now).

I can go ahead and adapt it as is to show the chart, but I think we should wait for further feedback before doing any more work, whether that is going with this block, or making a custom one.

helenb (approved):
Ready to go in

PR #129, opened 2024-11-13
Add back to top link at mobile, and tidy some bottom margins

helenb:
See https://jira.ons.gov.uk/browse/NWP-547
This pr adds the back to top link, for mobile only (easy to change to desktop too if we need to)

### How to review
Check the prototype pages at mobile to see the new back-to-top link in action.

### DIFF
--- a/ons_alpha/static_src/sass/components/design_system_overrides/_back-to-top.scss
+++ b/ons_alpha/static_src/sass/components/design_system_overrides/_back-to-top.scss
@@ -1,0 +1,11 @@
+@use 'config' as *;
+
+// overrides to ons back-to-top styles
+.ons-back-to-top {
+    margin-top: rem-sizing(64);
+
+    // hide at desktop
+    @include media-query('m') {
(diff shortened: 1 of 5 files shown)


ons_alpha/static_src/sass/components/design_system_overrides/_back-to-top.scss
```diff
@@ -0,0 +1,11 @@
+@use 'config' as *;
+
+// overrides to ons back-to-top styles
```
helenb, line 3: I experimented with wrapping the back-top-top link in a custom div that we could apply our own styling to, without overriding the design system styles. However, this wasn't possible as the js to make the link sticky when scrolling relies on the content being the previous element: https://github.com/ONSdigital/design-system/blob/1300be4d8f0f58c2a32b80a3bea85a8fd6ee82d6/src/components/back-to-top/back-to-top.js#L4

ababic (approved):
LGTM!

PR #133, opened 2024-11-14
Front end styling for the chart on the topic page

helenb:
- Builds on https://github.com/ONSdigital/dis-wagtail-alpha/pull/126 (note this branch targets that branch so changes are clear)
- Adds a custom front-end template for the featured document + chart block on the topic page
- Adds pym.js to the topic page
- Makes some tweaks to the back-end code to make the context vars simpler, as we're no longer using the document block macro template for this

### To consider / notes
- I think we could consider renaming the block to FeaturedChart but have left it as-is for now
- The overall heading is hard-coded as 'featured' - it looks from the design that this is intended instead of the featured document block - we could potentially simplify and combine these blocks into one.

### How to review
- Test with the relevant chart embed on the topic page - https://www.ons.gov.uk/visualisations/dvc3082/fig01_featured/index.html
- You can also now test the fallback image as datavis have updated the embed to include one.

### Screenshot

<details>
    <summary>Open sesame</summary>

Screenshot 2024-11-14 at 14 28 48

</details>

### DIFF
--- a/ons_alpha/core/blocks/featured_document.py
+++ b/ons_alpha/core/blocks/featured_document.py
@@ -77,6 +77,6 @@ class FeaturedDocumentWithChartBlock(FeaturedDocumentBlock):
     def get_context(self, value, parent_context=None):
         context = super().get_context(value, parent_context=parent_context)
         page = value["page"].specific
-        context["link_title"] = page.title
+        context["link_title"] = page.headline or page.title
         context["link_url"] = page.get_url(parent_context.get("request") if parent_context else None)
         return context
--- a/ons_alpha/core/blocks/featured_document.py
+++ b/ons_alpha/core/blocks/featured_document.py
@@ -58,8 +58,8 @@ class FeaturedDocumentBlock(blocks.StructBlock):
 
 
 class FeaturedDocumentWithChartBlock(FeaturedDocumentBlock):
-    chart_title = blocks.CharBlock(label=_("Chart title"), required=True)
     chart_url = blocks.URLBlock(label=_("Chart URL"), required=True)
+    chart_release_date = blocks.DateBlock(label=_("Release date"), required=True)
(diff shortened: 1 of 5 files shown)


ons_alpha/core/blocks/featured_document.py
```diff
@@ -58,8 +58,8 @@ def to_table_of_contents_items(self, value):  # pylint: disable=unused-argument
 
 
 class FeaturedDocumentWithChartBlock(FeaturedDocumentBlock):
-    chart_title = blocks.CharBlock(label=_("Chart title"), required=True)
```
ababic, line 61: @helenb I know it wasn't in the design, but surely we want to display a heading in our HTML before the embed? That's what this is for.

ons_alpha/core/blocks/featured_document.py
```diff
@@ -77,6 +84,11 @@ class Meta:
     def get_context(self, value, parent_context=None):
         context = super().get_context(value, parent_context=parent_context)
         page = value["page"].specific
-        context["link_title"] = page.headline or page.title
+        context["link_title"] = page.headline or page.display_title
```
ababic, line 87: What is `display_title`? Is that the admin-facing thing that uses the title of the most recent draft? If so, `page.title` is what we want here, as that's the official public-facing label.

helenb: See the change that Mebin made this morning: https://github.com/ONSdigital/dis-wagtail-alpha/pull/134

helenb: (this was after a conversation about how the bulletin page title should appear that we had yesterday in stand-up)

ababic: Okay, cool! Thanks for clarifying.

ababic (approved):
LGTM

PR #134, opened 2024-11-14
Update page titles

MebinAbraham:
Small tweak to accommodate:
- All page titles should be suffixed with `- Office for National Statistics` (except home?)
- Wherever the bulletin title is displayed, it should be prefixed with the series name.

### How to review
- Ensure page title includes the site name
- Bulletins page title should include the series name
- Previous releases listing should include the series name

### DIFF
--- a/ons_alpha/bulletins/models.py
+++ b/ons_alpha/bulletins/models.py
@@ -112,8 +112,12 @@ class BulletinPage(BundledPageMixin, RoutablePageMixin, BasePage):
     ]
 
     @property
+    def display_title(self):
+        return f"{self.get_parent().title}: {self.title}"
+
+    @property
     def full_title(self):
-        return self.headline.strip() or f"{self.get_parent().title}: {self.title}"
+        return self.headline.strip() or self.display_title
 
     @property
     def is_latest(self):
--- a/ons_alpha/jinja2/templates/base.html
+++ b/ons_alpha/jinja2/templates/base.html
@@ -90,11 +90,17 @@
     {% set body_class = '' %}
 {% endif %}
 
+{%- set page_title -%}
+    {%- with current_site=wagtail_site() -%}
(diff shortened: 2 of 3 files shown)


ons_alpha/bulletins/models.py
```diff
@@ -111,9 +111,13 @@ class BulletinPage(BundledPageMixin, RoutablePageMixin, BasePage):
         index.AutocompleteField("get_admin_display_title"),
     ]
 
+    @property
```
helenb, line 114: Query: would making this a @cached_property be more performant?

ons_alpha/jinja2/templates/base.html
```diff
@@ -90,11 +90,17 @@
     {% set body_class = '' %}
 {% endif %}
 
+{%- set page_title -%}
+    {%- with current_site=wagtail_site() -%}
+        {% if current_site and page.pk == current_site.root_page.pk and current_site.site_name %}{{ current_site.site_name }} - {% endif %}{% if page.seo_title %}{{ page.seo_title }}{% else %}{{ page.display_title | default(page.title) }}{% endif %}{% if current_site and page.pk != current_site.root_page.pk and current_site.site_name %} - {{ current_site.site_name }}{% endif %}
```
helenb, line 95: I wonder if we should give other page types a display_title in order to avoid having to check every time for its existence and then fall back on the page title?

helenb (approved):
Thanks @MebinAbraham I've tested these changes and happy they make sense. Just a couple of queries about performance - @ababic might have a more knowledgable opinion on these than me.

ons_alpha/jinja2/templates/base.html
MebinAbraham: There is not much we can do here; most times, it would be a title as it is mandatory. The default was a shorthand of doing another if statement which we do to check if they have an seo title instead.

ons_alpha/bulletins/models.py
MebinAbraham: It would be for computational-intensive methods, but this is quite simple hence negligible or worse here as `get_parent` also stores a cache itself.

helenb: Noted, thanks.

ons_alpha/jinja2/templates/base.html
helenb: OK, sounds good to leave it as is.

helenb:
Note, I'll update the title on the chart  on the topic page once this is merged.

PR #135, opened 2024-11-15
Corrections block - tweaks to spacing and improvements to screen reading

helenb:
- Ensures corrections block is flush with the header above
- Instead of using an after pseudo-element for 'hide content / show content', use 2 elements that are hidden / shown when the details block is open / closed. This allows us to hide them from the screen reader which announces that the summary is collapsed or open anyway. It also allows us to make the svg icons part of the markup, instead of background images, which makes the styling for focus state / high contrast mode etc much more simple.

### How to review
Check out a bulletin page and add some corrections or notices to review the changes

### Screenshots
<details>
<summary>Now look here</summary>

Corrections block now flush against the header

Screenshot 2024-11-15 at 09 59 45

</details>


ababic (approved):
Really nice improvement! Love it :)

PR #158, opened 2025-01-16
Update the banner wording and add a link

helenb:
See https://jira.ons.gov.uk/browse/NWP-568

### How to review
Test that the banner appears correctly and the link works as expected. It should open in a new tab.

### Screenshot

<details>
<summary>Open sesame</summary>
Screenshot 2025-01-16 at 08 43 48
</details>

### DIFF
--- a/ons_alpha/jinja2/templates/base_page.html
+++ b/ons_alpha/jinja2/templates/base_page.html
@@ -48,7 +48,7 @@
             <div class="ons-container">
                 {{
                 onsPanel({
-                "body": '<p>This page is part of a limited preview of the ONS website. We have intentionally made most links inactive for this preview.</p>',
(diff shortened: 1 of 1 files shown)


ons_alpha/jinja2/templates/base_page.html
```diff
@@ -34,23 +35,30 @@
 
 {% set languages=translation_urls() %}
 
-{#{% block header %}#}
-{#    {% secondary_nav %}#}
-{#    {% primary_nav %}#}
-{#{% endblock header %}#}
-
+{# fmt:off #}
```
ababic, line 38: @helenb This is typically added just before breaking into Jinja code (and enabled again right after). We still want formatter rules to apply to block content in general, so should probably stick with that convention

ababic (changes requested):

ons_alpha/jinja2/templates/base_page.html
helenb: As mentioned on slack it kept doing strange indenting *outside* the `{#fmt#}` tags when I tried using them closer in to the affected code.

ababic (approved):

PR #161, opened 2025-02-10
Add a new setting for the CPI page url

helenb:
- Adds a new link in the navigation menu to the CPI page
- Relies on a new environment variable called `ONS_CPI_PAGE_URL` - added to both [staging](https://dashboard.heroku.com/apps/ons-nov-prototype-staging/settings) and [production](https://dashboard.heroku.com/apps/ons-nov-prototype-production/settings) heroku environments. These will need updating once the relevant page is published on staging or production.
- If the env var is not set, the link will not appear in the navigation.

### How to review
- You can test by setting an environment variable in your `.env` file and then stopping and starting the container, for example `ONS_CPI_PAGE_URL="https://www.ons.gov.uk/economy/inflationandpriceindices/"`
- You should see a link in the navigation menu to this url on the "Inflation and price indices" nav item.

### DIFF
--- a/ons_alpha/jinja2/templates/base.html
+++ b/ons_alpha/jinja2/templates/base.html
@@ -294,7 +294,7 @@
                                 },
                                 {
                                     "text": "Inflation and price indices",
-                                    "url": False,
+                                    "url": CPI_PAGE_URL,
                                 },
                                 {
                                     "text": "National accounts",
--- a/ons_alpha/core/context_processors.py
+++ b/ons_alpha/core/context_processors.py
@@ -12,6 +12,7 @@ def global_vars(request):
         "LANGUAGE_CODE": settings.LANGUAGE_CODE,
         "IS_EXTERNAL_ENV": settings.IS_EXTERNAL_ENV,
         "TOPIC_PAGE_URL": settings.ONS_TOPIC_PAGE_URL,
+        "CPI_PAGE_URL": settings.ONS_CPI_PAGE_URL,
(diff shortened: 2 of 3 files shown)


ons_alpha/jinja2/templates/base.html
```diff
@@ -294,7 +294,7 @@
                                 },
                                 {
                                     "text": "Inflation and price indices",
-                                    "url": False,
```
RealOrangeOne, line 297: Question: Currently, if `$ONS_CPI_PAGE_URL` isn't defined, this will change from `False` to `""` - will that break anything? Should the default in `settings.py` be `False` to keep existing behaviour?

nimasmi: Good catch, though I don't think it makes a difference. They're rendered in https://github.com/ONSdigital/dis-wagtail-alpha/blob/november-prototype/ons_alpha/jinja2/component_overrides/header/_macro.njk#L200-L211 with the condition:

```jinja2
{% if child.url %}
```

I'm happy to leave this unchanged for expediency.

nimasmi (approved):
I approve.

I note the env variable is set, but with no value, for the time being. This is something to fix once confirmed on the ticket, and can be done after deployment if necessary.

ons_alpha/jinja2/templates/base.html
RealOrangeOne: Yep I'm happy with it as-is - falsey is falsey.

PR #164, opened 2025-02-12
NWP-584 Explore More StreamField block

nimasmi:

Ticket: https://jira.ons.gov.uk/browse/NWP-584

This adds an "Explore More" block. It is actually a variant implementation of the existing Design System [Document List](https://service-manual.ons.gov.uk/design-system/components/document-list) component.

### How to review
In the Wagtail admin, on a Bulletin page, add an Explore More block to the body.

<details>
<summary>Screenshots</summary>
Admin UI:

image

Rendered:
image

</details>

### DIFF
--- a/ons_alpha/core/blocks/__init__.py
+++ b/ons_alpha/core/blocks/__init__.py
@@ -1,6 +1,7 @@
 from .button_link import ButtonLinkBlock
 from .document_list import DocumentListBlock
 from .embeddable import DocumentBlock, DocumentsBlock, ImageBlock, ONSChartEmbedBlock, ONSEmbedBlock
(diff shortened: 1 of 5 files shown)


helenb:
I think (we should double check though) that this one should be for the topic page only. EDIT: I have double checked - definitely the topic page where this is needed.

nimasmi:
> I think (we should double check though) that this one should be for the topic page only. EDIT: I have double checked - definitely the topic page where this is needed.

Thanks for confirming. I have moved it to be usable on TopicPage only.

helenb (approved):
Tested and works well.

PR #160, opened 2025-02-04
Fix section markup for TOC

helenb:
### What is the context of this PR?

So far, only fixed for the topic page. Bulletin page also needs addressing.

### How to review
Describe the steps required to test the changes (include screenshots/screencasts if appropriate).

PR #167, opened 2025-02-14
fix spacing on section blocks on the bulletin page

helenb:
After updating the section markup in https://github.com/ONSdigital/dis-wagtail-alpha/pull/160, a further spacing fix was needed to ensure that the original spacing of streamfield blocks is followed.

### How to review
Check out this branch on a local build and make sure that each heading + content block is separated by 4 rems of spacing. If there is no contact details block on the page, the last streamfield block will not have a bottom margin set.

### DIFF
--- a/ons_alpha/jinja2/templates/pages/bulletin_page.html
+++ b/ons_alpha/jinja2/templates/pages/bulletin_page.html
@@ -107,7 +107,10 @@
             {% for block in page.body %}
                 {% if block.block_type == "heading" %}
                     {% if ns.has_section %} </section>{% endif %}
-                    <section id="{{ block.value|slugify }}">
+                    <section
+                        id="{{ block.value|slugify }}"
+                        class="{% if content_block.block_type !='heading' %}streamfield{% if last_flush %}streamfield--last-flush{% endif %}{% endif %}"
+                    >
                         <h2>{{ block.value }}</h2>
                         {% set ns.has_section = True %}
                 {% else %}
--- a/ons_alpha/jinja2/templates/pages/bulletin_page.html
+++ b/ons_alpha/jinja2/templates/pages/bulletin_page.html
@@ -109,12 +109,14 @@
                     {% if ns.has_section %} </section>{% endif %}
                     <section
                         id="{{ block.value|slugify }}"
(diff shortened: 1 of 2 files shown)


ons_alpha/jinja2/templates/pages/bulletin_page.html
```diff
@@ -107,11 +107,16 @@ <h1 class="ons-u-fs-3xl bulletin-header__heading">
...
+                    <section
+                        id="{{ block.value|slugify }}"
+                        class="streamfield {% if last_flush %}streamfield--last-flush{% endif %}"
+                    >
                         <h2>{{ block.value }}</h2>
                         {% set ns.has_section = True %}
                 {% else %}
-                    {% include_block block %}
+                    <div class="streamfield {% if loop.last %}streamfield--last-flush{% endif %}">
+                        {% include_block block %}
```
zerolab, line 118: doesn't this also come with a `<div class="streamfield">` wrapper?

zerolab: ah, it does not, but potentially has a different problem (if you have several SF blocks within the section)

or am I getting this wrong? sorry, Friday brain!

Screenshot 2025-02-14 at 11 16 55

helenb: Basically the div adds the extra spacing when there are several streamfield blocks within one section. It is possible that you get a nested div and section, both with a margin bottom, but in that situation the end result is still a margin bottom of 4 rem because it collapses.

There could still be an issue if an editor, say, added a heading above a document block - but they should not do that because the document block has it's own heading 2.

If we think it is a bit risky I would be happy to just revert the change to the section markup and not worry that the section highlighting doesn't quite work on the bulletin page - it feels more important for the spacing to be correct.

zerolab: understood. let's deploy this to staging and maybe get input from the editors? (or do they not have access to staging).

it would be safer to deploy the minimal changes, and follow this up next week (say on Tue?)

helenb:
I'm going to close this and revert to the minimal changes to fix things on the topic page.

PR #176, opened 2025-03-28
Update streamfield fields for Headline Figures and Explore More

MaciekBaron:
This PR addresses NWP-595 and NWP-596. It introduces an image field to the Headline Figures component, and a page/external link field to the Explore More component.

### How to review
* Checkout branch
* Create a topic or bulletin page
* Add a Headline Figures block
* Check that an image can be added
* Add an Explore More block
* Check that both a page and an external link can be added

### Follow-up tasks
The styling needs to be adjusted once we receive the images.

### DIFF
--- a/ons_alpha/core/blocks/explore_more.py
+++ b/ons_alpha/core/blocks/explore_more.py
@@ -1,7 +1,8 @@
+from django.core.exceptions import ValidationError
 from django.utils.html import format_html, strip_tags
 from django.utils.text import slugify
 from django.utils.translation import gettext as _
-from wagtail.blocks import CharBlock, ListBlock, StructBlock, TextBlock
+from wagtail.blocks import CharBlock, ListBlock, PageChooserBlock, StructBlock, TextBlock, URLBlock
 from wagtail.images.blocks import ImageChooserBlock
 
 
@@ -9,6 +10,16 @@ class ExploreMoreListItemBlock(StructBlock):
     title = CharBlock(label=_("Title"), max_length=200, required=True)
     thumbnail = ImageChooserBlock(label=_("Thumbnail"), required=True)
     description = TextBlock(label=_("Description"), max_length=300, required=True)
+    page = PageChooserBlock(
+        label=_("Page"),
+        required=False,
+        help_text=_("Select a page to link to. This will be used as the URL."),
+    )
+    external_link = URLBlock(
+        label=_("External link"),
+        required=False,
+        help_text=_("Enter an external URL to link to."),
+    )
 
     class Meta:
         icon = "list-ul"
@@ -17,6 +28,9 @@ class ExploreMoreListItemBlock(StructBlock):
     def clean(self, value):
         value = super().clean(value)
         value["description"] = strip_tags(value["description"])
+        # Ensure only one of page or external_link is provided
+        if value["page"] and value["external_link"]:
+            raise ValidationError(_("You cannot provide both a page and an external link at the same time."))
         return value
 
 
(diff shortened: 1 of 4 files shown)


ons_alpha/jinja2/templates/components/streamfield/headline_figures_block.html
```diff
@@ -12,6 +12,12 @@ <h3 class="headline-figures__title">
                         </a>
                     {% endif %}
                 </h3>
+                {% if figure.image %}
+                    <figure class="headline-figures__image-container">
+                        {{ srcset_image(figure.image, "format-png|width-{278,556}", class="headline-figures__image") }}
```
MaciekBaron, line 17: I chose `format-png` as the example images show diagrams with text, and a lossless format would be more appropriate here to preserve the content.

ons_alpha/core/blocks/explore_more.py
```diff
@@ -17,6 +28,9 @@ class Meta:
     def clean(self, value):
         value = super().clean(value)
         value["description"] = strip_tags(value["description"])
+        # Ensure only one of page or external_link is provided
+        if value["page"] and value["external_link"]:
+            raise ValidationError(_("You cannot provide both a page and an external link at the same time."))
```
nimasmi, line 33: **Thought** (non-blocking, depending how urgent this work is): you can make these appear by the fields in question if you like. See https://docs.wagtail.org/en/latest/advanced_topics/streamfield_validation.html#controlling-where-error-messages-are-rendered

```suggestion
            raise StructBlockValidationError(block_errors={
                "external_link": ValidationError(
                    _("You cannot provide both a page and an external link at the same time.")
                )
            })
```

ons_alpha/core/blocks/explore_more.py
```diff
@@ -45,6 +59,7 @@ def get_context(self, value, parent_context=None):
             document = {
                 "title": {
                     "text": item["title"],
+                    "url": item["page"].url if item["page"] else item["external_link"],
```
nimasmi, line 62: **Question**: …the latter of which could be an empty string, thus handling the no-link case?

ons_alpha/static_src/sass/components/_headline-figures.scss
```diff
@@ -84,6 +84,10 @@
         }
     }
 
+    &__image-container {
+        margin: 0 0 1rem 0;
```
nimasmi, line 88: **Praise**: Yep, these look close enough for now (once I ran `npm run build` :flushed:)

nimasmi:
This looks good to me. There's an optional suggestion, depending on your preferred editor messaging experience and time available.

nimasmi (approved):

ons_alpha/core/blocks/explore_more.py
MaciekBaron: Thanks, I was thinking about that, however this technically relates to two fields so I thought it would be more appropriate to raise it for the whole block. Happy to make this change though!

ons_alpha/core/blocks/explore_more.py
MaciekBaron: Yep, the macro template just checks if `url` is truthy so it will work with an empty string.

https://github.com/ONSdigital/dis-wagtail-alpha/blob/d1ae70f2f6b12df5e221c9f6607b5beedc8ab099/ons_alpha/jinja2/component_overrides/document-list/_macro.njk#L22-L27

ons_alpha/core/blocks/explore_more.py
nimasmi: If this were a 'real' project, I'd say we needed to reconcile this UX with that at https://github.com/ONSdigital/dis-wagtail-alpha/blob/c76d34683bc0a1219c57cc8024395186ad78080f/ons_alpha/core/blocks/document_list.py#L35-L38 (added yesterday!), for consistency. But as it is, this project is not under heavy use by inexperienced editors, so I'm fine with the discrepancy, too.

